PR #24, opened 2020-07-08
add exists_ok parameter to Files.create_dir

orlof:
Added support to exists_ok parameter in Files.create_dir.
This changes the default behaviour to not raise an FileExistsError if directory already exists.

### DIFF
--- a/packages/main/src/RPA/FileSystem.py
+++ b/packages/main/src/RPA/FileSystem.py
@@ -283,36 +283,36 @@ class FileSystem:
         with open(path, "ab") as fd:
             fd.write(content)
 
-    def create_directory(self, path, parents=False, exists_ok=True):
+    def create_directory(self, path, parents=False, exist_ok=True):
         """Creates a directory and (optionally) non-existing parent directories.
 
         :param path:        path to new directory
         :param parents:     create missing parent directories
-        :param exists_ok:   FileExistsError is ignored if path is an existing directory
+        :param exist_ok:    FileExistsError is ignored if directory already exists
         """
-        Path(path).mkdir(parents=parents, exist_ok=exists_ok)
+        Path(path).mkdir(parents=parents, exist_ok=exist_ok)
 
-    def remove_file(self, path, force=False):
+    def remove_file(self, path, missing_ok=True):
         """Removes the given file.
 
-        :param path:    path to the file to remove
-        :param force:   ignore non-existent files
+        :param path:        path to the file to remove
+        :param missing_ok:  ignore non-existent file
         """
         try:
             Path(path).unlink()
         except FileNotFoundError:
-            if not force:
+            if not missing_ok:
                 raise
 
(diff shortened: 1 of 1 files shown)


osrjv:
In other keywords we've used the argument `force` to mirror command-line terminology, but `exist_ok` would be more in line with the pathlib API. We should probably decide which way is better. The terms `exist_ok` and `missing_ok` are maybe more correct since no forcing of anything is really happening.

orlof:
`force` would sound counter intuitive. Argument defines whether it is ok not to do anything, if directory already exists.

orlof:
Do you want me to modify the remove_file(s) keywords to take in `missing_ok=True`, instead of the `force`? That would be more compatible with python 3.8 pathlib.

osrjv:
I guess we should go with `missing_ok`, but this change will be backwards incompatible. Note that we can't actually use that argument for `unlink` since we need to support older Python versions. Also change the new argument for creating directories to be `exist_ok` (without plural) to actually match pathlib.

orlof:
Update done.

packages/main/src/RPA/FileSystem.py
```diff
@@ -283,35 +283,36 @@ def append_to_binary_file(self, path, content):
...
 
-        :param paths:   paths to files to be removed
-        :param force:   ignore non-existent files
+        :param paths:       paths to files to be removed
+        :param missing_ok:  ignore non-existent files
         """
         # TODO: glob support
         for path in paths:
-            self.remove_file(path, force)
+            self.remove_file(path, missing_ok=missing_ok)
```
janipalsamaki, line 315: Is this a typo? `missing_ok=missing_ok`

janipalsamaki: Never mind! :P

orlof:
> Approved the conceptual change. However, this changes the behavior in a way that might cause unexpected things for users (who might expect a raised error, or that there is a `force` argument).

In philosophical level, I think that some python developers may expect pathlib -style behaviour, but for RFW users exception handling is less convenient. I believe that in majority of the cases in RPA domain file remove is supposed to ensure that file does not exist (or mkdir is supposed to ensure that dir exists) instead of actually ensuring that the operation if fulfilled.

Originally `force` was a wrong term to describe that not-doing-something is ok.

osrjv:
In the future we really want to stick with semver and this would be a major version change, no way around it. However I don't think we want to do those sorts of releases too often. Maybe we should collect breaking changes into a separate branch to batch them up a bit, and add a deprecation notice to the current release about the upcoming changes. This will add some maintenance overhead, but we can live with it, I think.

osrjv:
The new `exist_ok` argument is now merged into master and will be part of the next release. The replacement of `force` with `missing_ok` has been cherry-picked into `release/3.0` pending the next major release.

Thanks for your contribution(s)!

PR #35, opened 2020-08-06
correct error message

Noordsestern:
message did not state when SMTP server does not support TLS and logged confusing warning instead.


Noordsestern:
Closed issue, since i cannot sign CLA in name of the company. Contributing in private instead.

PR #36, opened 2020-08-06
correct error message

Noordsestern:
original warned about SMTP server not supporting SMTP when it only did not support TLS but worked fine anyway.


osrjv:
Thanks for the PR! The error reason was obviously wrong (and quite confusing)

PR #43, opened 2020-08-14
Include read_from_cells based on write_to_cells

JooooooooCo:
Guys, I've include "read_from_cells" in my personal library, based in write_to_cells... It work in For loop very well, like UiPath uses Read Cell function....


mikahanninen:
Thank you for adding this! Nice addition to the library.

PR #49, opened 2020-09-24
Add mss as dependency and use for screenshotting

xylix:
BREAKING CHANGES: removes ´save_format´ argument from take_screenshot and `crop_image` functions / Robot keywords

Using mss improves screenshot delay / latency by almost 90% (on my machine changes from 2-2.5 seconds to 150-250ms).

I created a new test, which reveals (at least on my machine) that find by template finds the screenshot at a different location from the one specified when the screenshot was taken. This is a problem with matching templates, but the speed is heavily improved. I'll try to run the tests on windows as well to see if that works nicely.

TODO:
- [x] Remove now unnecessary pyscreenshot dependency
- [x] test on Windows

Manually tested that `Take Screenshot` and `Crop Image` work on windows and macOS. Also added a new test, but it has some issues with an offset in "find by template" (the coordinate space for the find_by function might be flipped or something, but that's out of scope for this PR)


packages/main/src/RPA/Images.py
```diff
@@ -163,35 +163,38 @@ def __init__(self):
...
+
         region = to_region(region)
 
-        if region is not None:
-            image = ImageGrab.grab(bbox=region.as_tuple())
-        else:
-            image = ImageGrab.grab()
+        with mss.mss() as sct:
+            if region is not None:
+                image = sct.grab(region.as_tuple())
```
osrjv, line 177: Is the origin the top-left of the main monitor or the whole virtual screen?

packages/main/src/RPA/Images.py
```diff
@@ -200,7 +203,8 @@ def crop_image(self, image, region, filename=None, save_format="PNG"):
         image.load()
 
         if filename:
-            image.save(filename, save_format)
+            # Suffix isn't created automatically here
+            image.save(filename + ".png", "PNG")
```
osrjv, line 207: should this also use pathlib?

packages/main/src/RPA/Images.py
```diff
@@ -243,7 +247,7 @@ def find_template_in_image(
         if not matches:
             raise ImageNotFoundError("No matches for given template")
 
-        # Convert region coördinates back to full-size coördinates
```
osrjv, line 246: Oh no, you removed my favourite stupid spelling: https://en.wiktionary.org/wiki/co%C3%B6rdinate#English

osrjv (changes requested):
Talking about conventions, check that `make lint` passes. We auto-format everything through black and I think some of the line removals will error out.

packages/main/src/RPA/Images.py
xylix: Oh my :D I'll undo the change

packages/main/src/RPA/Images.py
xylix: Yeah that's probably a more stable solution

packages/main/src/RPA/Images.py
xylix: I am pretty sure the origin is the top-left of the whole virtual screen. mss docs were a bit lean on this so I'm unsure, will try to test

PR #50, opened 2020-09-24
Feature/database library

mikahanninen:


packages/main/src/RPA/Database.py
```diff
@@ -1,23 +1,527 @@
+import importlib
 import logging
-from DatabaseLibrary import DatabaseLibrary
+import sys
 
 from RPA.Tables import Table
 
+try:
+    import ConfigParser
```
osrjv, line 8: This is a python2 compatibility thing, I don't think we need it, right?

packages/main/src/RPA/Database.py
```diff
@@ -1,23 +1,527 @@
...
+        host,
+        port,
+        charset,
+        config_file: str,
+    ):
+        config = ConfigParser.ConfigParser()
+        config.read([config_file])
+
+        self.configuration = {}
+        self.module_name = module_name or (
```
osrjv, line 36: The configparser api provides the option for fallbacks, so it looks like you could write this pattern like this:
`self.module_name = module_name or config.get("default", "module_name", fallback=None)`

packages/main/src/RPA/Database.py
```diff
@@ -1,23 +1,527 @@
...
+        )
+        self.configuration["charset"] = charset or (
+            config.get("default", "charset")
+            if config.has_option("default", "charset")
+            else None
+        )
+        return self.module_name, self.configuration
+
+    def get(self, param, default=None):
+        return (
```
osrjv, line 74: This too can be `self.configuration.get(param, default)`

packages/main/src/RPA/Database.py
```diff
@@ -1,23 +1,527 @@
...
+        """
+        params = params or []
+        cur = None
+        try:
+            if self.db_api_module_name == "cx_Oracle":
+                cur = self._dbconnection.cursor()
+            else:
+                cur = self._dbconnection.cursor(as_dict=False)
+            PY3K = sys.version_info >= (3, 0)
+            if not PY3K:
```
osrjv, line 292: All of these python3 version checks should probably be removed at some point.

packages/main/src/RPA/Database.py
```diff
@@ -1,23 +1,527 @@
...
+                    if not valid:
+                        raise AssertionError(
+                            "Query assertion %s failed. Facts: %s"
+                            % (assertion, available_locals)
+                        )
+                if as_table:
+                    return Table(rows, columns)
+                return rows
+            else:
+                if result is not None:
```
osrjv, line 452: Is this correct? It seems to be committing twice in some situtations.

packages/main/src/RPA/Database.py
```diff
@@ -1,23 +1,527 @@
...
+                retVal.append(row)
+            if not sanstran:
+                self._dbconnection.commit()
+            return retVal
+        finally:
+            if cur:
+                if not sanstran:
+                    self._dbconnection.rollback()
+
+    def description(self, table):
```
osrjv, line 307: This keyword could be (ironically) a bit more descriptive

packages/main/src/RPA/Database.py
```diff
@@ -1,23 +1,527 @@
...
+
+    def __init__(self) -> None:
         self.logger = logging.getLogger(__name__)
-        DatabaseLibrary.__init__(self, *args, **kwargs)
+        self._dbconnection = None
+        self.db_api_module_name = None
+        self.config = Configuration()
+
+    # pylint: disable=R0915
+    def connect_to_database(  # noqa: C901
```
osrjv, line 115: I know this keyword is mostly copied from upstream but splitting it up would make a bit more readable/extendable. Not high prio though.

docs/source/libraries/database/index.rst
```diff
@@ -12,10 +12,14 @@ Description
 
 `Database` is a library for handling different database operations.
 
-The library extends `robotframework-databaselibrary`_.
+This can allow you to query your database after an action has been made to verify the results.
```
osrjv, line 15: Is this the main use case for us? Shouldn't it be about making those actions instead?

osrjv (commented):
Some random comments

PR #53, opened 2020-09-28
Add new changes to screenshot and crop keywords to releasenotes

xylix:


docs/source/releasenotes.rst
```diff
@@ -13,6 +13,10 @@ Upcoming release
 - Library **PDF**: Keywords ``Template HTML To PDF`` and ``HTML To PDF`` will now
   create directory structure and overwrite existing file by default.
 
+- Library **RPA.Images**: Remove ``Save Format`` option from ``Take Screenshot``
```
mikahanninen, line 16: remove RPA prefix from the library name .... that applies to all libraries and is not needed in the release note

xylix: Oh right whoops.

xylix: Fixed

PR #56, opened 2020-09-30
Feature/dialog library

mikahanninen:


packages/main/src/RPA/Dialogs.py
```diff
@@ -0,0 +1,428 @@
...
+                # etag = response.headers.get("ETag")
+                if response.status_code == 200:
+                    try:
+                        response_json = response.json()
+                        break
+                    except JSONDecodeError:
+                        pass
+                elif response.status_code != 304:
+                    # back off if the server is throwing errors
+                    time.sleep(60)
```
osrjv, line 488: 60 seconds seems a bit excessive

packages/main/src/RPA/Dialogs.py
```diff
@@ -0,0 +1,428 @@
...
+            )
+        return formhtml
+
+    def get_title(self, item):
+        return f"<h3>{item['value']}</h3>"
+
+    def get_text(self, item):
+        return f"<p>{item['value']}</p>"
+
+    def create_form(self, message):
```
osrjv, line 141: This HTML generator should probably be separate, and I think we have bs4 already as a dependency which would have made this a bit easier. But no need actions needed for now, just keep in mind if we keep working on this (a lot) more.

packages/main/src/RPA/Dialogs.py
```diff
@@ -0,0 +1,428 @@
...
+            br.open_available_browser(f"{self.server_address}/form.html")
+            br.set_window_size(window_width, window_height)
+
+            headers = {"Prefer": "wait=120"}
+            response_json = None
+            # etag = None
+            while True:
+                # if etag:
+                #    headers['If-None-Match'] = etag
+                headers["If-None-Match"] = "2434432243"
```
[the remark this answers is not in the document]
mikahanninen: debug stuff for the long polling ;) I will remove if I can't make long polling work

packages/main/src/RPA/Dialogs.py
```diff
@@ -0,0 +1,428 @@
...
+    ROBOT_LIBRARY_SCOPE = "GLOBAL"
+
+    def __init__(self):
+        self.logger = logging.getLogger(__name__)
+        self.server_address = None
+        self.server = None
+        self.workdir = None
+        self.custom_form = None
+
+    def start_attended_server(self, port=8105):
```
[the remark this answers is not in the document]
mikahanninen: yeah I change it to private method

packages/main/src/RPA/Dialogs.py
mikahanninen: I was thinking to make it separate but did not want to go there at this stage. Good point on bs4 - did not notice that it is part of our dependencies.

PR #57, opened 2020-10-05
Move finding or erroring on a fail from Browser.py to locators.py

xylix:
Found some locator finding functionality in Browser.py that I will need to reuse with the new image aliases. If using the internal logger is a problem for this I can also fix that, but to me this seems like a reasonable refactor.

### DIFF
--- a/packages/core/src/RPA/core/locators.py
+++ b/packages/core/src/RPA/core/locators.py
@@ -35,6 +35,31 @@ def open_stream(obj, *args, **kwargs):
             obj.close()
 
 
+def find_by_name_or_error(path, criteria):
+    db = LocatorsDatabase(path)
+    if not Path(db.path).exists():
+        db.logger.warning("File does not exist: %s", db.path)
+
+    db.load()
+    if db.error:
+        error_msg, error_args = db.error
+        raise ValueError(error_msg % error_args)
+
+    entry = db.find_by_name(criteria)
+    if not entry:
+        raise ValueError(f"Unknown locator alias: {criteria}")
+
+    if entry["type"] != "browser":
+        raise ValidationError(f"Not a browser locator: {criteria}")
+
+    locator = "{prefix}:{criteria}".format(
+        prefix=entry["strategy"], criteria=entry["value"]
+    )
+
+    db.logger.info("%s is an alias for %s", criteria, locator)
+    return locator
+
+
 class ValidationError(ValueError):
     """Validation error from malformed database or locator entry."""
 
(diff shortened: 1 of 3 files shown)


packages/core/src/RPA/core/locators.py
```diff
@@ -34,6 +35,31 @@ def open_stream(obj, *args, **kwargs):
...
 
+def find_by_name_or_error(path, criteria):
+    db = LocatorsDatabase(path)
+    if not Path(db.path).exists():
+        db.logger.warning("File does not exist: %s", db.path)
+
+    db.load()
+    if db.error:
+        error_msg, error_args = db.error
+        raise ValueError(error_msg % error_args)
```
osrjv, line 46: I think this should be ValidationError

packages/core/src/RPA/core/locators.py
```diff
@@ -34,6 +35,31 @@ def open_stream(obj, *args, **kwargs):
...
+
+    entry = db.find_by_name(criteria)
+    if not entry:
+        raise ValueError(f"Unknown locator alias: {criteria}")
+
+    if entry["type"] != "browser":
```
osrjv, line 52: I don't think this should be here? The library using this can decide if it supports the type or not.

packages/core/src/RPA/core/locators.py
```diff
@@ -34,6 +35,31 @@ def open_stream(obj, *args, **kwargs):
...
+        raise ValidationError(f"Not a browser locator: {criteria}")
```
osrjv, line 53: ...and this shouldn't be ValidationError

packages/core/src/RPA/core/locators.py
```diff
@@ -34,6 +35,31 @@ def open_stream(obj, *args, **kwargs):
             obj.close()
 
 
+def find_by_name_or_error(path, criteria):
```
osrjv, line 38: maybe just `find_by_name` would be enough

osrjv (changes requested):

packages/core/src/RPA/core/locators.py
xylix: Yes absolutely.

packages/core/src/RPA/core/locators.py
xylix: I was thinking of this ValidationError as something that should come up when a locator is invalid in format / data. They won't pop up here because db.load() resets validation errors:

```
 def load(self):
        """Deserialize database from file."""
        try:
            with open_stream(self.path, "r") as fd:
                data = json.load(fd)

            self._validate_all(data)
            self._locators = data
            self.reset_error()
```

packages/core/src/RPA/core/locators.py
xylix: Well the LocatorDatabase exposes a `find_by_name` method that doesn't fit the need of `Browser` module or the needs image aliases will soon have.

Maybe `load_by_name` could work, since it would represent that we actually load the db from disk when this function is called?

packages/core/src/RPA/core/locators.py
xylix: Fixed at https://github.com/robocorp/rpaframework/pull/58/files

packages/core/src/RPA/core/locators.py
osrjv: It resets if nothing before it raised an exception, but sets an error in the except branch

packages/core/src/RPA/core/locators.py
osrjv: That would work, I think. Some other "loading" can also happen behind the scenes, like parsing images

packages/core/src/RPA/core/locators.py
xylix: Oh right of course. I'll change that to ValidaitonError then

packages/core/src/RPA/core/locators.py
xylix: Changed at https://github.com/robocorp/rpaframework/pull/58/files

packages/core/src/RPA/core/locators.py
xylix: Changed at https://github.com/robocorp/rpaframework/pull/58/files

PR #60, opened 2020-10-05
Load locator by name Browser.py changes

xylix:
Requires a release with changes from #59 before can be merged


osrjv:
Merged manually

PR #61, opened 2020-10-06
Feature: Work item files

osrjv:
This branch adds support for files in work items. No changes to existing keywords, but requires cloud support until it can be deployed.

### DIFF
--- a/docs/source/libraries/robocloud_items/index.rst
+++ b/docs/source/libraries/robocloud_items/index.rst
@@ -36,10 +36,14 @@ overriden manually while developing an activity.
 Item structure
 ==============
 
-A work item's payload is JSON and allows storing anything that is
+A work item's data payload is JSON and allows storing anything that is
 serializable. This library creates an object with the key 'variables'
 that contains key-value pairs of a variable name and its contents.
 These variables can be exposed to the Robot Framework task to be used directly.
+
+In addition to the data section, a work item can also contain files,
+which are stored by default in Robocorp's cloud. Adding and using
+files with work items requires no additional setup from the user.
 
 Workflow
 ========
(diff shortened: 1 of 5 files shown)


packages/core/src/RPA/core/locators.py
```diff
@@ -5,20 +5,15 @@
...
+    return os.path.join(dirname, filename)
+
 
-    if PROJECT_PATH_ENV in os.environ:
-        # locators.json is found at root of project, use environment var
-        # if available
-        return os.path.join(os.environ[PROJECT_PATH_ENV], DEFAULT_DATABASE_NAME)
-    else:
-        return DEFAULT_DATABASE_NAME
+DEFAULT_DATABASE = default_locators_path()
```
xylix, line 16: Good improvements in readability

docs/source/libraries/robocloud_items/index.rst
```diff
@@ -36,11 +36,15 @@ overriden manually while developing an activity.
 Item structure
 ==============
 
-A work item's payload is JSON and allows storing anything that is
+A work item's data payload is JSON and allows storing anything that is
 serializable. This library creates an object with the key 'variables'
 that contains key-value pairs of a variable name and its contents.
 These variables can be exposed to the Robot Framework task to be used directly.
 
+Parallel to the data section, a work item can also contain files,
```
xylix, line 44: Parallel might be an unclear term here? Maybe just a boring "In addition ..." could work?

docs/source/libraries/robocloud_items/index.rst
```diff
@@ -78,13 +82,14 @@ the suite starts.
...
-        Log   Using variables from workspace ${workspace} for user ${user_id}
-
+    Save variables to Robocloud
+        Add work item file    orders.xlsx
+        Set work item variables    user=Dude    mail=address@company.com
+        Save work item
 
-In the following example the work item is modified locally and then saved
-back to Robocloud.
+Later in the process inside a different robot, we can use previously saved
```
xylix, line 90: Do work items have any sort of namespacing / should you be careful to not collide names if you are running many robots in parallel?

docs/source/libraries/robocloud_items/index.rst
osrjv: A bit clunky, yes. Will fix.

docs/source/libraries/robocloud_items/index.rst
osrjv: Currently the work item data payload is just any JSON and the robot can freely modify it. It is however reserved for the execution and there will be no parallel access. Namespacing might or might not come when we start pushing reusable/shareable robots.

xylix: Okay clear enough. Then the documentation from this PR should be sufficient

PR #70, opened 2020-11-02
Feature Request: Add 'value' to a text input

mdp:
Would it be possible to add a 'value' to the text inputs to allow for presenting a default value to the user. We'd like to prompt the user for input, but we want to be able to set the value in the text input. This would be similar to the current state of vanilla 'Dialogs' where 'Get Value From User Input' accepts a default value.


mikahanninen:
Thank you for the pull request. Seems like something that can be added to the library.

mikahanninen (approved):

PR #71, opened 2020-11-04
Lock numpy version 1.19.3 because of Windows 10 2004 breakage

xylix:
https://developercommunity.visualstudio.com/content/problem/1207405/fmod-after-an-update-to-windows-2004-is-causing-a.html


mikahanninen:
Thank you for the heads-up 👍  We will pin it for the next release.

PR #77, opened 2020-11-10
Kerkko/lab 1207 click aliastemplate uses wrong main changes

xylix:
Requires rpaframework-core 2.2.* in pypi

### DIFF
--- a/packages/main/src/RPA/Desktop/keywords/finder.py
+++ b/packages/main/src/RPA/Desktop/keywords/finder.py
@@ -10,6 +10,8 @@ from RPA.core.locators import (
     parse_locator,
 )
 
+from RPA.Desktop.keywords.screen import _get_displays
+
 try:
     from RPA.recognition import templates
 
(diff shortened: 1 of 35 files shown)


packages/main/src/RPA/Desktop/keywords/finder.py
```diff
@@ -10,6 +10,8 @@
     parse_locator,
 )
 
+from RPA.Desktop.keywords.screen import _get_displays, _region_from_mss_monitor
```
osrjv, line 13: if these are used outside the `screen` module, they probably should not prefixed with underscores

packages/main/src/RPA/Desktop/keywords/finder.py
```diff
@@ -33,6 +35,50 @@ def __init__(self, ctx):
         else:
             self.confidence = None
 
+    def get_matches(self, locator: ImageTemplate) -> List[Union[Point, Region]]:
```
osrjv, line 38: this could be a bit more descriptive, like `find_templates`

packages/main/src/RPA/Desktop/keywords/finder.py
```diff
@@ -33,6 +35,50 @@ def __init__(self, ctx):
...
+        """Internal helper method for getting image template matches in all displays
+        and returning them as Points or Regions, scaled to accomodate macOS HiDPI
+        """
+
+        def get_scaled_matches(screenshot, locator, display):
+            """ Internal helper function for finding matches on a single screen """
+            self.logger.info("Matching with confidence of %.1f", confidence)
```
osrjv, line 45: this could be outside the looping so we don't log the same thing for every monitor

osrjv:
Merged manually

PR #78, opened 2020-11-11
browser: add highlight_elements

mchece:
Highlight functionality inspired by Robocode Lab and robotframework-browser (Playwright) library.

### DIFF
--- a/packages/main/src/RPA/Browser.py
+++ b/packages/main/src/RPA/Browser.py
@@ -1626,3 +1626,69 @@ class Browser(SeleniumLibrary):
                 "download.directory_upgrade": True,
                 "download.prompt_for_download": False,
             }
+
+    @keyword
+    def highlight_elements(
+        self,
+        locator: str,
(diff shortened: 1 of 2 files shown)


packages/main/src/RPA/Browser.py
```diff
@@ -1626,3 +1626,69 @@ def set_download_directory(
...
+            self.driver.execute_script(script)
+
+        def highlight():
+            script = "".join(
+                f'arguments[{idx}].setAttribute("{attribute_name}", "");'
+                for idx in range(len(elements))
+            )
+            self.driver.execute_script(script, *elements)
+
+        inject_style()
```
osrjv, line 1678: If there's only one caller right after the definition, might as well not have these as nested functions

mchece: I agree on `clear_all_highlights` - on the above one I think it separates the two steps nicely.

PR #79, opened 2020-11-16
Add default filename and indexing for Take Screenshot

xylix:
This is a subset of the automatic filenaming that is done in browser and Selenium. Adds option to have `{index}` as part of filename string to get that replaced with an index number like below

rpa-desktop-screenshot-1.png
rpa-desktop-screenshot-2.png


osrjv:
Cherry-picked into master with other changes

PR #107, opened 2020-12-14
desktop: update docstrings

mchece:
Update documentation for Desktop library.

### DIFF
--- a/packages/main/src/RPA/Desktop/__init__.py
+++ b/packages/main/src/RPA/Desktop/__init__.py
@@ -26,6 +26,8 @@ class Desktop(DynamicCore):
     - Mouse and keyboard input emulation
     - Starting and stopping applications
     - Finding elements through image template matching
+    - Finding elements through OCR matching
+    - Scraping text from given regions
     - Taking screenshots
     - Clipboard management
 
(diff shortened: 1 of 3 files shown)


packages/main/src/RPA/Desktop/__init__.py
```diff
@@ -44,28 +46,45 @@ class Desktop(DynamicCore):
...
-        Move mouse    offset:200,0
+        Move mouse  image:%{ROBOT_ROOT}/logo.png
+        Move mouse  offset:200,0
         Click
 
+        Click       point:50,100
+
+        Click       region:20,20,100,30
+
+        Click       ocr:Create New Account
```
osrjv, line 77: broken with current master. when #110 is merged, this would work as `ocr:"Create New Account"`

packages/main/src/RPA/Desktop/keywords/screen.py
```diff
@@ -202,6 +213,19 @@ def move_region(self, region: Region, left: int, top: int) -> Region:
...
+        Usage examples:
+
+        .. code-block:: robotframework
+
+            ${region}=          Find Element  ocr:Net Assets
+            ${moved_region}=    Move Region  ${region}  500  0
+
+        .. code-block:: python
+
+            region = desktop.find_element("ocr:Net Assets")
```
osrjv, line 226: This also should be quoted, like the rest of the ocr-with-whitespace-stuff

osrjv (commented):
I think this needs a section on the new syntax, but I think that can be done separately.

PR #112, opened 2020-12-16
Create workflow for pip nightly installs

xylix:


osrjv:
Like discussed, probably needs a matrix for (some) previous major versions, but I'm not sure where/how to get that info robustly. Probably through the PyPI API and a few lines of python/bash script.

PR #124, opened 2021-01-14
Feature/selenium pdf downloader

mchece:


packages/main/src/RPA/Browser/Selenium.py
```diff
@@ -1696,6 +1698,68 @@ def clear_all_highlights(self):
...
+            "printBackground": True,
+            "preferCSSPageSize": True,
+        }
+
+        try:
+            output_dir = BuiltIn().get_variable_value("${OUTPUT_DIR}", "output")
+        except robot.libraries.BuiltIn.RobotNotRunningError:
+            output_dir = "output"
+        default_output = f"{output_dir}/out.pdf"
+        output_path = output_path if output_path else default_output
```
osrjv, line 1754: this can also be said like this: `output_path = output_path or default_output`

packages/main/src/RPA/Browser/Selenium.py
```diff
@@ -1696,6 +1698,68 @@ def clear_all_highlights(self):
...
+
+        params = params if params else default_params
```
osrjv, line 1756: ditto

packages/main/src/RPA/Browser/Selenium.py
```diff
@@ -1696,6 +1698,68 @@ def clear_all_highlights(self):
...
+            {
+                "landscape": False,
+                "displayHeaderFooter": False,
+                "printBackground": True,
+                "preferCSSPageSize": True,
+            }
+        """
+        if "chrom" not in self.driver.name:
+            self.logger.warning("PDF printing works only with Chrome/Chromium")
+            return
```
osrjv, line 1726: If this can't actually do the thing it promises, should it just throw an exception?

packages/main/src/RPA/Browser/Selenium.py
mchece: :man_facepalming:

osrjv:
Not sure how I missed this earlier, but I'm a bit on the fence about the the first `source` argument for the keyword. Is there some reason to have it like that instead of printing the current page? As it is, it doesn't just print the given source page but also switches the entire driver there for future keywords.

osrjv:
Took the liberty of removing the argument, can still be reverted if I misunderstood something.

mchece:
Sure. Understand the problem there, just remove it for now.

PR #125, opened 2021-01-14
pdf: return textboxes with get_text_from_pdf and more locators to get…

mikahanninen:


packages/main/src/RPA/PDF.py
```diff
@@ -1098,48 +1133,70 @@ def get_value_from_anchor(
...
+        if (
+            direction == "right"
+            and self._is_within_tolerance(item.top, top)
+            and item.left >= right
+        ):
+            direction_ok = True
+        elif (
+            direction == "left"
+            and self._is_within_tolerance(item.top, top)
+            and item.right < left
```
osrjv, line 1184: should this also be inclusive like `right`?

packages/main/src/RPA/PDF.py
```diff
@@ -60,6 +60,14 @@ def iterable_items_to_int(bbox):
 logging.getLogger("pdfminer").setLevel(logging.WARNING)
 
 
+class TargetObject:
```
osrjv, line 63: consider using a dataclass

packages/main/src/RPA/PDF.py
```diff
@@ -1157,12 +1214,24 @@ def _is_match_on_vertical(self, direction, item, strict, regexp):
                 return item
         return None
 
+    def _is_match_in_box(self, item):
```
osrjv, line 1217: `RPA.core.geometry` has primitives from points and regions with common helper methods. not necessary maybe to refactor here, but as a future note.

packages/main/src/RPA/PDF.py
```diff
@@ -1098,48 +1133,70 @@ def get_value_from_anchor(
...
+        return item if match else None
 
     def _is_match_on_vertical(self, direction, item, strict, regexp):
         (left, bottom, right, top) = self.anchor_element.bbox
         text = None
-        if (direction == "bottom" and item.top < bottom) or (
-            direction == "top" and item.bottom > top
+        direction_down = direction in ["bottom", "down"]
+        direction_up = direction in ["top", "up"]
+        if (direction_down and item.top < bottom) or (
```
osrjv, line 1198: and should these also be inclusive like the horizontal comparisons? not sure if technically the elements can even overlap.

packages/main/src/RPA/PDF.py
osrjv: I mean `<=` instead of `<`

PR #129, opened 2021-01-20
Docs: JSON library (+ TODO documentation)

osrjv:
This pull request adds some examples and further explanations to the existing JSON library.

In addition, it adds a small tool for collating all TODO comments in code and adding it to the contributing section of Sphinx.

Preview of libdoc output: http://whispering-work.surge.sh/json.html

### DIFF
--- a/packages/main/src/RPA/JSON.py
+++ b/packages/main/src/RPA/JSON.py
@@ -1,6 +1,6 @@
 import json
 import logging
-from typing import Any
+from typing import Any, Dict, List, Union
 
 from jsonpath_ng import Index, Fields
 from jsonpath_ng.ext import parse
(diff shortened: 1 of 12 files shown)


packages/main/src/RPA/JSON.py
```diff
@@ -9,13 +9,98 @@
...
+    In the simplest case JSONPath can replace nested access:
+
+    .. code-block:: robotframework
+
+       # First order of first client, with direct dictionary access
+       ${value}=    Set variable    ${json}["clients"][0]["orders"][0]
+
+       # JSONPath access
+       ${value}=    Get value from JSON    ${json}    $.clients[0].orders[0]
+
```
mchece, line 65: This got me thinking... I like Python doctest style usage examples
```>>> sum([1,2,3])
6
```
Does Robot Framework documentation have a tradition in these kinds of examples?

packages/main/src/RPA/JSON.py
```diff
@@ -24,21 +109,43 @@ def __init__(self):
...
 
-        :param filename: path to file
-        :return: json as a dictionary object
         """
         self.logger.info("Loading JSON from file: %s", filename)
         with open(filename) as json_file:
             return json.load(json_file)
 
     @keyword("Save JSON to file")
     def save_json_to_file(self, doc: Any, filename: str) -> None:
```
mchece, line 134: I would hint the type like 
```
doc: Union[str, dict]
```
but not sure if in the scope of this PR.

packages/main/src/RPA/JSON.py
```diff
@@ -24,21 +109,43 @@ def __init__(self):
...
-        """Save JSON object into a file.
+        """Save a JSON object, or a string containg a JSON value, into a file.
```
mchece, line 135: Is it a "JSON object" or an ordinary Python dictionary...?

packages/main/src/RPA/JSON.py
```diff
@@ -24,21 +109,43 @@ def __init__(self):
 
     @keyword("Load JSON from file")
     def load_json_from_file(self, filename: str) -> Any:
```
mchece, line 114: I would hint the possible return types instead of `Any`

packages/main/src/RPA/JSON.py
osrjv: Not exactly like that. I've tried to display it in keyword form (at least in some other examples). There's a wish/requirement that the examples can be copy-pasted easily, and quite often keywords are more about side-effects than return values. Maybe it could be as a comment below it, if relevant?

packages/main/src/RPA/JSON.py
osrjv: You can of course define type aliases, but you can also enjoy this long conversation: https://github.com/python/typing/issues/182

It gets... less useful when you support all JSON things like plain scalar values as root objects.

packages/main/src/RPA/JSON.py
osrjv: See above about JSONs. Technically "JSON serializable object" I guess?

packages/main/src/RPA/JSON.py
osrjv: Maybe we can go with: `JSON = Union[str, int, float, bool, None, Dict[str, Any], List[Any]]`

mchece: Yeah, I was afraid it would lead to something like the linked discussion.

PR #138, opened 2021-02-03
PDF library refactor, package split

mchece:
Refactored old PDF code from `main` package into its own `PDF` package. Added tests to cover the most simple cases.


packages/pdf/pyproject.toml
```diff
@@ -0,0 +1,61 @@
+[tool.poetry]
+name = "rpaframework-pdf"
+version = "0.2.0"
+description = "PDF utilities used by RPA Framework"
```
osrjv, line 4: Probably the PDF library of RPA Framework and not just utilities.

packages/pdf/pyproject.toml
```diff
@@ -0,0 +1,61 @@
...
+]
+
+packages = [
+	{ include = "RPA", from = "src" }
+]
+
+[tool.poetry.dependencies]
+python = "^3.6"
+rpaframework-core = "^5.2.0"
+rpaframework-recognition = "^0.6.0"
```
osrjv, line 37: Do we really need recognition non-optionally?

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,510 @@
...
+        return self._output_directory
+
+    @output_directory.setter
+    def output_directory(self, path: str):
+        self.output_directory = Path(path)
+
+    def __del__(self):
+        self.close_all_pdf_documents()
+
+    def close_all_pdf_documents(self) -> None:
```
osrjv, line 44: This seems to be a duplicate of the function in the root library.

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,510 @@
...
+
+    @property
+    def output_directory(self):
+        return self._output_directory
+
+    @output_directory.setter
+    def output_directory(self, path: str):
+        self.output_directory = Path(path)
+
+    def __del__(self):
```
osrjv, line 41: Python will clean up file handles when the references go out of scope, and in general __del__ is a bit unpredictable. Would just remove it.

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,510 @@
...
+        """
+        if source_pdf is None:
+            raise ValueError("Source PDF is missing")
+        if str(source_pdf) in self.ctx.fileobjects.keys():
+            raise ValueError(
+                "PDF file is already open. Please close it before opening again."
+            )
+        self.ctx.active_pdf = str(source_pdf)
+        self.ctx.active_fileobject = open(source_pdf, "rb")
+        self.ctx.active_fields = None
```
osrjv, line 73: should the active document be some sort of container for these? keeping the path and handle in sync seems to be a bit of a hassle.

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,510 @@
...
+        """Use HTML content to generate PDF file.
+
+        :param content: HTML content
+        :param target_pdf: filepath where to save PDF document
+        :param variables: dictionary of variables to fill into template, defaults to {}
+        :param create_dirs: directory structure is created if it is missing,
+         default `True`
+        :param exists_ok: file is overwritten if it exists, default `True`
+        """
+        required_param([content, target_pdf], "html_to_pdf")
```
osrjv, line 95: Is there a reason for this function to exist? Why not have the parameters not have default values in the first place?

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,510 @@
...
+        for key, value in variables.items():
+            html = html.replace("{{" + key + "}}", str(value))
+
+        default_output = Path(self.output_directory / "html2pdf.pdf")
+        output_filepath = Path(target_pdf) if target_pdf else default_output
+
+        self._write_html_to_pdf(html, output_filepath, create_dirs, exists_ok)
+
+    def _write_html_to_pdf(self, html, output_path, create_dirs, exists_ok):
+        # TODO: is this needed?
```
osrjv, line 109: Figure out TODO, obviously. IMO it could just always overwrite and always create missing parent dirs.

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,510 @@
...
+    def _add_pages(self, pages: int = 1) -> None:
+        """Adds pages into PDF documents.
+
+        :param pages: number of pages to add, defaults to 1
+        """
+        for _ in range(int(pages)):
+            self.fpdf.add_page()
+
+    @keyword
+    def get_info(self, source_pdf: str = None) -> dict:
```
osrjv, line 133: Kinda vague naming.. could it be something like `get document info`?

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,510 @@
...
+            to `output_directory`
+        :param pages: page numbers to extract from PDF (numbers start from 0)
+        :param clockwise: directorion that page will be rotated to, default True
+        :param angle: number of degrees to rotate, default 90
+        """
+        self.switch_to_pdf_document(source_pdf)
+        reader = PyPDF2.PdfFileReader(self.ctx.active_fileobject)
+        writer = PyPDF2.PdfFileWriter()
+
+        default_output = Path(self.output_directory / "rotated.pdf")
```
osrjv, line 304: Does the workflow make sense where each keyword call on a document stores it into a new file? Wouldn't you want to do a bunch of ops on the file in-memory and then as the final step store it on disk?

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,510 @@
...
+                else:
+                    source_page.rotateCounterClockwise(int(angle))
+            else:
+                source_page = reader.getPage(int(page))
+            writer.addPage(source_page)
+        with open(str(output_filepath), "wb") as f:
+            writer.write(f)
+
+    @keyword
+    def pdf_encrypt(
```
osrjv, line 325: sometimes keyword names use `document`, something just `pdf`, and sometimes `pdf document`. maybe we should just choose one.

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,510 @@
...
+        :param replace: used to replace `text`
+        """
+        if source_pdf or self.rpa_pdf_document is None:
+            self.ctx.convert(source_pdf)
+        for _, page in self.rpa_pdf_document.get_pages().items():
+            for _, textbox in page.get_textboxes().items():
+                if textbox.text == text:
+                    textbox.text = replace
+                    return
+        self.logger.warning("Did not find any matching text")
```
osrjv, line 410: Or just throw error?

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,510 @@
...
+        return False
+
+    @keyword
+    def replace_textbox_text(self, text: str, replace: str, source_pdf: str = None):
+        """Replace text content of a textbox with something else in the PDF.
+
+        :param text: this text will be replaced
+        :param replace: used to replace `text`
+        """
+        if source_pdf or self.rpa_pdf_document is None:
```
osrjv, line 403: The source document handling seems to be a bit inconsistent. Sometimes it's switched automatically to the active document and sometimes not.

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,510 @@
...
+        writer = PyPDF2.PdfFileWriter()
+        pdf = FPDF()
+        pdf.add_page()
+        reader = PyPDF2.PdfFileReader(source)
+        mediabox = reader.getPage(0).mediaBox
+        im = Image.open(imagefile)
+        width, height = im.size
+        max_width = int(float(mediabox.getWidth()) * coverage)
+        max_height = int(float(mediabox.getHeight()) * coverage)
+        if width > max_width:
```
osrjv, line 454: This if/else can still result in an image that is larger than the document.

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,510 @@
...
+
+    @keyword
+    def save_pdf(
+        self, source: str = None, target: str = None, custom_reader: PyPDF2.PdfFileReader = None
+    ):
+        """Save current over itself or to `target_pdf`.
+
+        :param source: filepath to source PDF. If not given, the active fileobject is used.
+        :param target: filepath to target PDF
+        :param custom_reader: a modified PDF reader.
```
osrjv, line 482: I have no idea what this is for.

packages/pdf/src/RPA/PDF/keywords/model.py
```diff
@@ -0,0 +1,585 @@
...
+)
+
+
+def iterable_items_to_int(bbox) -> list:
+    if bbox is None:
+        return list()
+    return list(map(int, bbox))
+
+
+class RpaFigure:
```
osrjv, line 43: I'm for some reason really bothered by the naming of these model classes. Why are they RPA-specific primitives?

packages/pdf/src/RPA/PDF/utils.py
```diff
@@ -0,0 +1,52 @@
+import os
+import platform
+import time
+
+
+class Buffer:
```
osrjv, line 6: Is there a reason this was copied over from Desktop? I don't see it used anywhere

packages/pdf/src/RPA/PDF/keywords/model.py
```diff
@@ -0,0 +1,585 @@
...
+
+    def write_footer(self):
+        self.write("</pages>\n")
+
+    def write_text(self, text):
+        if self.stripcontrol:
+            text = self.CONTROL.sub("", text)
+        self.write(enc(text))
+
+    def receive_layout(self, ltpage):  # noqa: C901 pylint: disable=R0915
```
osrjv, line 234: This function sure is... something.

osrjv (commented):
I think the next step would be to generate the keyword docs from this and make sure that:
- The keywords are named consistently
- The arguments are named consistently
- Keywords have a consistent return value
- The documentation describes what the keyword _actually_ does

After that make sure:
- Active document handling is the same (for all source document arguments)
- Destination document handling is the same (for all destination document arguments), including creating parent dirs etc.
- Possibly apply changes to documents consistently in-memory instead of sometimes writing it out to a file?

packages/pdf/src/RPA/PDF/keywords/model.py
mchece: This is some kind of pdfminer magic method... need documenting

packages/pdf/src/RPA/PDF/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # Register keyword libraries to LibCore
+        libraries = [
+            DocumentKeywords(self),
+            FinderKeywords(self),
+            ModelKeywords(self),
+        ]
+        super().__init__(libraries)
+
+        # TODO: how to use this RPA.main keyword library
```
osrjv, line 50: This log manipulation stuff should be moved to core, if it's required in multiple places

packages/pdf/src/RPA/PDF/__init__.py
```diff
@@ -0,0 +1,52 @@
+import logging
+from robotlibcore import DynamicCore
+
+from RPA.PDF.keywords import DocumentKeywords, FinderKeywords, ModelKeywords
+
+
+class PDF(DynamicCore):
+    """`PDF` is a library for managing PDF documents.
```
osrjv, line 8: Something a bit more about the features and use-cases would be nice.

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,898 @@
...
+        """
+        output_path = Path(output_path) if output_path else self.default_output
+        self._write_html_to_pdf(content, output_path)
+
+    def _write_html_to_pdf(self, html: str, output_path: str) -> None:
+        self.logger.info("Writing output to file %s", output_path)
+        self.fpdf.add_page()
+        self.fpdf.write_html(html)
+        self.fpdf.output(name=output_path)
+        # self.__init__()  # TODO: what should happen here exactly?
```
osrjv, line 228: Remove?

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,898 @@
...
+        self.fpdf = PDF()
```
osrjv, line 229: if a new instance is being created for each invocation of `_write_html_to_pdf()`, maybe fpdf could just be a local var in the function and not stored in the class

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,898 @@
...
+        """
+        # TODO: don't save to a new file every time
+        self.switch_to_pdf(source_path)
+        reader = PyPDF2.PdfFileReader(self.ctx.active_pdf_document.fileobject)
+        writer = PyPDF2.PdfFileWriter()
+
+        output_filepath = Path(output_path) if output_path else self.default_output
+
+        if not isinstance(pages, list):
+            pagelist = [pages]
```
osrjv, line 596: This page number handling is different than the previous keywords.

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,898 @@
...
+                )
+
+        :param image_path: filepath to image file to add into PDF
+        :param source: filepath to source, if not given add image to currently
+            active PDF
+        :param output_path: filepath of target PDF
+        :param coverage: [description], defaults to 0.2
+        :raises ValueError: [description]
+        """
+        if source_path is None and self.ctx.active_pdf_document.fileobject.path:
```
osrjv, line 827: Should this also use the switch to pdf method?

packages/pdf/src/RPA/PDF/keywords/document.py
```diff
@@ -0,0 +1,898 @@
...
+
+        return width, height
+
+    @keyword
+    def save_pdf(
+        self,
+        output_path: str,
+        reader: PyPDF2.PdfFileReader,
+    ):
+        """Save the contents of a PyPDF2 reader to a new file.
```
osrjv, line 883: Does this mean something to the user?

packages/pdf/src/RPA/PDF/keywords/model.py
```diff
@@ -0,0 +1,733 @@
...
+        if save:
+            self.update_field_values(None, None, self.ctx.active_pdf_document.fields)
+
+    @keyword
+    def update_field_values(
+        self, source_path: str = None, output_path: str = None, newvals: dict = None
+    ) -> None:
+        """Update field values in PDF if it has fields.
+
+        This is used internally in the library.
```
osrjv, line 598: Only internally or externally too?

packages/pdf/src/RPA/PDF/keywords/model.py
```diff
@@ -0,0 +1,733 @@
...
+    def close(self) -> Document:
+        self.write_footer()
+        return self.active_pdf_document
+
+
+class ModelKeywords(LibraryContext):
+    """Keywords for converting PDF document into specific RPA object model"""
+
+    @keyword
+    def convert(self, source_path: str = None) -> None:
```
osrjv, line 376: Should this be visible to users or done automatically? It also has a fairly generic name that very likely overlaps with user keywords or other libraries

osrjv:
Merged manually to `release/8.0` branch

PR #149, opened 2021-03-11
Feature/secrets/update secret

mchece:
Add keyword `Set Secret`


osrjv:
Changed argument wording a bit to be less tied to the current implementation, but otherwise LGTM

PR #192, opened 2021-06-21
jab: initial library commit

mikahanninen:
Java Access Library "beta" version


packages/main/src/RPA/JavaAccessBridge.py
```diff
@@ -0,0 +1,520 @@
...
+
+    def _click_element_middle(self, element):
+        self.logger.info("Element click coordinates")
+        left, top, right, bottom = self._get_scaled_coordinates(element)
+        if left == -1 or top == -1:
+            raise AttributeError("Can't click on negative coordinates")
+        middle_x = int((left + right) / 2)
+        middle_y = int((top + bottom) / 2)
+        point = f"point:{middle_x},{middle_y}"
+        Desktop().click(point)
```
osrjv, line 451: The `Desktop` library will accept geometry primitives defined in `RPA.core.geometry`, i.e. `Region` and `Point` instances. No need to convert things into a string first. You can probably do just something like this:
```
region = Region.from_size(element.left, element.top, element.width, element.height)
region.scale(self.scale_factor)
Desktop().click(region.center)
```

mmokko (commented):
Looks good, probably would be better to use monotonic time comparison instead of os time.

packages/main/src/RPA/JavaAccessBridge.py
```diff
@@ -0,0 +1,520 @@
...
+    ):
+        """Selects Java application window as target for the automation
+
+        :param title: application window title
+        :param bring_foreground: if application is brought to foreground or not
+        :param timeout: selection timeout
+        """
+        window_found = False
+        interval = float(0.5)
+        end_time = time.time() + float(timeout)
```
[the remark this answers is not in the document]
osrjv: might be a case of polishing a turd.... even all internal RF timeouts use non-monotonic time.

packages/main/src/RPA/JavaAccessBridge.py
```diff
@@ -0,0 +1,520 @@
...
+    def _initialize(self):
+        pipe = queue.Queue()
+        self.pumper_thread = threading.Thread(
+            target=self._pump_background, daemon=True, args=[pipe]
+        )
+        self.pumper_thread.start()
+        self.jab_wrapper = pipe.get()
+        if not self.jab_wrapper:
+            raise Exception("Failed to initialize Java Access Bridge Wrapper")
+        time.sleep(1)
```
osrjv, line 165: Why the sleep? It'll block a second for a each execution that imports this library

packages/main/src/RPA/JavaAccessBridge.py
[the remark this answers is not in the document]
mmokko: Still no reason for us not to use monotonic time. They can fix the RF side at some point and then we would have to fix our libraries. Not saying we should do this for all libraries, but at least the new ones we create? Libraries also can be used as pure python code.

packages/main/src/RPA/JavaAccessBridge.py
```diff
@@ -0,0 +1,520 @@
...
+    def __init__(self):
+        self.logger = logging.getLogger(__name__)
+        self.version_printed = False
+        self.jab_wrapper = None
+        self.context_info_tree = None
+        self.pumper_thread = None
+        self._initialize()
+        self.refresh_counter = 1
+        self.display_scale_factor = (
+            ctypes.windll.shcore.GetScaleFactorForDevice(0) / 100
```
osrjv, line 153: I think we need to return early from here since calling stuff from windll won't work on other platforms, and it'd be nice if imports worked on other platforms. At least to get libdoc stuff generated.

packages/main/src/RPA/JavaAccessBridge.py
```diff
@@ -0,0 +1,520 @@
...
+        )
+
+    def _initialize(self):
+        pipe = queue.Queue()
+        self.pumper_thread = threading.Thread(
+            target=self._pump_background, daemon=True, args=[pipe]
+        )
+        self.pumper_thread.start()
+        self.jab_wrapper = pipe.get()
```
osrjv, line 162: This should probably be called with some sort of timeout value. Blocking forever in case of an error until this is killed externally is maybe not the best. Even RF timeouts won't trigger here, I think?

packages/main/src/RPA/JavaAccessBridge.py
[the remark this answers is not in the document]
osrjv: sure, just saying. using `time.monotonic` is probably better if it doesn't need to be compatible with `time.time`. we have 15ish places in rpaframework where this would have to be fixed as well.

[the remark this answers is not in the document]
osrjv: I can't see an official spec that `time.sleep` uses monotonic time, but the cpython implementation happens to be

packages/main/tests/robot/test_java.robot
```diff
@@ -0,0 +1,60 @@
...
+    Click Element    role:push button and name:Clear
+
+*** Keywords ***
+Task setup actions
+    Select Window    Chat Frame
+    Clear chat frame
+    #Application Refresh
+
+*** Tasks ***
+Click element
```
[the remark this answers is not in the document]
mikahanninen: That **Click Element** is a task name, not a best name I agree. I will rename those task names.

packages/main/src/RPA/JavaAccessBridge.py
mikahanninen: Seems like nice way of using Region. I will add this as an library TODO ... no time to change and retest everything this week.

packages/main/src/RPA/JavaAccessBridge.py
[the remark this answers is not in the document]
mikahanninen: maybe something that could be changed cross rpaframework at a later date ?

packages/main/src/RPA/JavaAccessBridge.py
mikahanninen: sleep has been now removed

packages/main/src/RPA/JavaAccessBridge.py
mikahanninen: I have organized imports so that build is green on all platforms and library init will make WARN log when using non-Windows platform

packages/main/src/RPA/JavaAccessBridge.py
mikahanninen: @mmokko I think jab wrapper package handles now so that the thread is not blocked forever ?

mmokko: Yeah the error will get it out, but still a good point. Maybe add a ten second timeout for example, just loading the library shouldn't take more than tens of milliseconds.

packages/main/src/RPA/JavaAccessBridge.py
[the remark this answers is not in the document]
mmokko: Yeah, maybe over engineering on my part, just gotten used to use the monotonic time API with different languages.

packages/main/src/RPA/JavaAccessBridge.py
mikahanninen: I added 10 seconds timeout to pipe.get()

PR #222, opened 2021-09-10
Date picker in RPA.Dialogs

### LINKED ISSUE #219: RPA.Dialogs date picker
Add the possibility of picking up a date from the dialog UI

- [x] Support internally a date picking React element (in **robocorp-dialog** dependency)
- [ ] `Add Date Input` keyword in RF displaying and collecting data from the element

### Limitations
- Can't easily pick a date far away from the current year (or provided default)
- Can't input date as text directly in the entry (read-only value property)

cmin764:
Closes #219 

- Adds support for date picking widget under `RPA.Dialogs`
- Robot and unit-tests
- See [limitations](https://github.com/robocorp/rpaframework/issues/219)

Tested with this:
```robotframework
Get Birthdate
    Add icon      Success
    Add Date Input    birthdate    26/04/1993
    ${result} =    Run dialog    title=Success
    Log To Console    ${result.birthdate}
```

### DIFF
--- a/packages/dialogs/src/RPA/Dialogs/library.py
+++ b/packages/dialogs/src/RPA/Dialogs/library.py
@@ -702,32 +702,28 @@ class Dialogs:
         self,
         name: str,
         default: Optional[str] = None,
-        label: Optional[str] = None,
     ) -> None:
         """Add a date input element
 
         :param name:    Name of result field
         :param default: The default date
-        :param label:   Label for input field (not supported yet)
 
         Displays a date input widget. The selection the user makes will be available
         in the ``name`` field of the result.
         The ``default`` argument can be a pre-set date in DD/MM/YYYY format, otherwise
         the current date is used.
-        A custom ``label`` text can also be added.
 
         Example:
 
         .. code-block:: robotframework
 
-            Add heading     Enter your birthdate
-            Add Date Input
-            ...    name=birthdate
-            ${result}=      Run dialog
+            Add heading       Enter your birthdate
+            Add Date Input    birthdate    default=26/04/1993
+            ${result} =       Run dialog
             Log    User birthdate should be: ${result.birthdate}
         """
 
-        # TODO(cmiN): Be flexible on date formats. (provide it as parameter)
+        # TODO(cmin764): Be flexible on date formats. (provide it as parameter)
         py_date_format = "%d/%m/%Y"
         js_date_format = "dd/MM/yyyy"
         if default:
@@ -745,7 +741,6 @@ class Dialogs:
             "name": str(name),
             "format": js_date_format,
             "default": optional_str(default),
-            "label": optional_str(label),
         }
 
         self.add_element(element)
--- a/packages/dialogs/src/RPA/Dialogs/library.py
+++ b/packages/dialogs/src/RPA/Dialogs/library.py
@@ -708,7 +708,7 @@ class Dialogs:
 
         :param name:    Name of result field
         :param default: The default date
-        :param label:   Label for input field
+        :param label:   Label for input field (not supported yet)
 
         Displays a date input widget. The selection the user makes will be available
         in the ``name`` field of the result.
--- a/packages/dialogs/src/RPA/Dialogs/library.py
+++ b/packages/dialogs/src/RPA/Dialogs/library.py
@@ -4,7 +4,7 @@ import glob
 import logging
 import mimetypes
 import time
-from datetime import datetime
+from datetime import date, datetime
 from pathlib import Path
 from typing import Dict, List, Optional, Union, Any, Generator
 
@@ -701,7 +701,7 @@ class Dialogs:
     def add_date_input(
         self,
         name: str,
-        default: Optional[str] = None,
+        default: Optional[Union[date, str]] = None,
     ) -> None:
         """Add a date input element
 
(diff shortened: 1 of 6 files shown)


packages/dialogs/src/RPA/Dialogs/library.py
```diff
@@ -696,6 +697,53 @@ def add_drop_down(
...
+        if default:
+            try:
+                datetime.strptime(default, format)
+            except Exception as exc:
+                raise ValueError(f"Invalid default date with value {default!r}") from exc
+        else:
+            default = datetime.utcnow().strftime(format)
+
+        element = {
+            "type": "input-datepicker",
```
cmin764, line 739: We're blocked here by this other robocorp-dialogs [PR](https://github.com/robocorp/robocorp-dialog/pull/1).

packages/dialogs/src/RPA/Dialogs/library.py
```diff
@@ -696,6 +697,53 @@ def add_drop_down(
 
         self.add_element(element)
 
+    @keyword("Add date-picker", tags=["input"])
```
mikahanninen, line 700: I think the name for the  keyword should actually be **Add Date Input** to follow naming convention within current library implementation.

mikahanninen (commented):
It would be good to show at least as "todo" comments in the code that the keyword still requires support for "value", "min" and "max" for the date input field.

cmin764:
> show at least as "todo" comments in the code that the keyword still requires support for "value", "min" and "max" for the date input field.

Agree, but I feel this wouldn't make sense anymore if we use the WIP React [datepicker](https://robocorp-design-system-jq08uln2s-robocorp-dev.vercel.app/component/datepicker) (usage in [PR](https://github.com/robocorp/robocorp-dialog/pull/1)), right? As of the following:
- `value` is handled internally (picks up `default`'s value if provided)
- `min` & `max` not supported yet by the design system

----

Still, I can add some TODOs if we plan to add extra support in the future, but not sure yet what to write :).

packages/dialogs/src/RPA/Dialogs/library.py
cmin764: Not blocked anymore.

cmin764:
_Access rights required_

After merge, I should be enabled to publish the new `0.4.0` version to PyPI using my [user](https://pypi.org/user/cmiN/).

```console
UploadError

HTTP Error 403: The user 'cmiN' isn't allowed to upload to project 'rpaframework-dialogs'. See https://pypi.org/help/#project-name for more information.
```

packages/dialogs/src/RPA/Dialogs/library.py
```diff
@@ -696,6 +697,59 @@ def add_drop_down(
 
         self.add_element(element)
 
+    @keyword("Add Date Input", tags=["input"])
+    def add_date_input(
+        self,
+        name: str,
+        default: Optional[str] = None,
+        label: Optional[str] = None,
```
osrjv, line 705: I would just drop this from the public API if it isn't supported now. Optional arguments can always be added later.

osrjv:
Seems to work quite nice!

There's something here that I didn't think of earlier. The current result is a string in the DD/MM/YYYY format, right? What would be the way for a user to actually use that as a date in their robot? Should we post-process that somewhere into a python datetime object?

cmin764:
> Seems to work quite nice!
> 
> There's something here that I didn't think of earlier. The current result is a string in the DD/MM/YYYY format, right? What would be the way for a user to actually use that as a date in their robot? Should we post-process that somewhere into a python datetime object?

Yes, the date is returned as a string and I guess the dev could use some date [conversion](https://robocorp.com/docs/libraries/built-in/datetime/keywords#convert-date) mechanism to get an object out of it.
```robotframework
Library           DateTime
...
Convert date test
    ${date_obj} =    Convert Date    26/04/1993    date_format=%d/%m/%Y    result_format=datetime
    Log To Console    ${date_obj}

----
> Get Birthdate                                                         1993-04-26 00:00:00
```
I believe devs got used to expect strings out of the inputs, but if you feel is better to return the object directly, we can do it. If "yes", any idea on how to convert this at its best? I believe we can add a check for the "input-datepicker" element under _packages/dialogs/src/RPA/Dialogs/dialog.py_ so we can convert it into an object from a string when such element is occured.

osrjv:
I guess strings make more sense when we implement the `format` argument, so they can match it with what the system is expecting. There are some exceptions in the results already, like checkboxes being actual booleans.

The parsing would probably be in `src/RPA/Dialogs/dialog.py:143` where the results are parsed and compared to the given elements. However, maybe it can be left to the user. An example in the keyword would be nice though, e.g. how to split year/mont/day from the result.

cmin764:
> I guess strings make more sense when we implement the `format` argument, so they can match it with what the system is expecting. There are some exceptions in the results already, like checkboxes being actual booleans.
> 
> The parsing would probably be in `src/RPA/Dialogs/dialog.py:143` where the results are parsed and compared to the given elements. However, maybe it can be left to the user. An example in the keyword would be nice though, e.g. how to split year/mont/day from the result.

Now the keyword returns a `date` object and accepts both `date[time]` and `str` as `default`. Tested manually too with:
```robotframework
Get Birthdate docstring
    Add heading       Enter your birthdate
    Add Date Input    birthdate    default=26/04/1993
    ${result} =       Run dialog
    Log To Console    User birthdate year should be: ${result.birthdate.year}
```

PR #234, opened 2021-09-23
In/Out Work Item env var path support

cmin764:
Adds support for reading & writing local dev work items from these env vars:
- `RPA_INPUT_WORKITEM_PATH`
- `RPA_OUTPUT_WORKITEM_PATH`

### ToDo
- [x] make Windows & TS/VSCode paths work correctly
- [x] unit & robot tests
- [x] review, accept/refactor, release **rpaframework-core** `6.5.0` (bugfix on `6.5.1`)
- [ ] squash-merge, ~publish **rpaframework** `11.3.0`~ (tag **rpaframework-core** `6.5.1`)


mmokko:
Looks good 👍, have to leave for the day, but I'll test this tomorrow morning with the plugin.

cmin764:
> tomorrow morning with the plugin.

Thanks, I'll have to reproduce and fix the Windows related paths bug too and we're done :).

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -288,21 +288,27 @@ class FileAdapter(BaseAdapter):
...
-    * RPA_WORKITEMS_PATH:   Path to work items database file
+    * RPA_INPUT_WORKITEM_PATH:   Path to work items database file
     """
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        path = required_env("RPA_WORKITEMS_PATH")
+        old_path = os.getenv("RPA_WORKITEMS_PATH", UNDEFINED_VAR)
+        if old_path is not UNDEFINED_VAR:
+            logging.warning("Work items old path style usage detected (deprecated)")
```
cmin764, line 306: Nit: Maybe issuing this warning once through `warnings.warn`? @mmokko

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -326,6 +332,22 @@ def get_input(self) -> str:
...
+    @property
+    def output_path(self):
+        if not self._output_path:
+            # This is usually set once per loaded input work item.
+            new_path = os.getenv("RPA_OUTPUT_WORKITEM_PATH")
+            if new_path:
+                self._output_path = resolve_path(new_path)
+                self._output_path.parent.mkdir(parents=True, exist_ok=True)
+            else:
+                logging.warning(
```
cmin764, line 344: See comment [above](https://github.com/robocorp/rpaframework/pull/234/files#r714855479).

packages/main/tests/python/test_robocorp_workitems.py
```diff
@@ -462,16 +462,30 @@ def test_custom_root(self, adapter):
...
+            with open(items_in, "w") as fd:
                 json.dump(ITEMS_JSON, fd)
             with open(os.path.join(datadir, "file.txt"), "w") as fd:
                 fd.write("some mock content")
 
-            monkeypatch.setenv("RPA_WORKITEMS_PATH", items)
+            yield items_in, items_out
+
+    @pytest.fixture(
+        params=[
```
cmin764, line 480: Runs all the tests with both old and new env vars so the behaviour is kept.

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -396,6 +418,7 @@ def remove_file(self, item_id: str, name: str):
 
         path = files[name]
         logging.info("Would remove file: %s", path)
+        # Note that the file doesn't get removed from disk as well.
```
cmin764, line 421: Check if this is right.

mmokko (approved):
Looks good to me. Tested with vscode using different path options, and got expected results 👍 .

packages/core/src/RPA/core/logger.py
```diff
@@ -11,6 +12,14 @@
 from RPA.core.helpers import required_param
 
 
+logging.captureWarnings(True)
+
+
+def deprecation(message):
```
cmin764, line 18: @osrjv Is this the right place to introduce such utility? (that I foresee to frequently use)

packages/core/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "rpaframework-core"
-version = "6.4.0"
+version = "6.5.0"
```
cmin764, line 3: I tested locally with a wheel, but in this PR, should I split this change in another PR? (along with all the other `core` changes)

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -288,21 +289,27 @@ class FileAdapter(BaseAdapter):
...
-    * RPA_WORKITEMS_PATH:   Path to work items database file
+    * RPA_INPUT_WORKITEM_PATH:   Path to work items database file
     """
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        path = required_env("RPA_WORKITEMS_PATH")
+        old_path = os.getenv("RPA_WORKITEMS_PATH", UNDEFINED_VAR)
+        if old_path is not UNDEFINED_VAR:
+            deprecation("Work items load - Old path style usage detected")
```
cmin764, line 307: Don't forget to put here the links to docs. (so user knows how to use the new style)

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -288,21 +289,27 @@ class FileAdapter(BaseAdapter):
...
+    with the suffix ``<filename>.output.json``. If the output path is provided by an
+    env var explicitly, then the file will be saved with the provided path and name.
 
     Reads and writes all work item files from/to the same parent
     folder as the given input database.
 
     Required environment variables:
 
-    * RPA_WORKITEMS_PATH:   Path to work items database file
+    * RPA_INPUT_WORKITEM_PATH:   Path to work items database file
```
cmin764, line 300: How about specifying the `RPA_OUTPUT_WORKITEM_PATH ` too, but this time under a new section of "Optional environment variables:" so we can link this doc in the `deprecation()` warnings.

packages/core/src/RPA/core/logger.py
osrjv: Seems like a good enough place

packages/core/src/RPA/core/logger.py
```diff
@@ -11,6 +12,14 @@
 from RPA.core.helpers import required_param
 
 
+logging.captureWarnings(True)
```
osrjv, line 15: I wonder if this could have some unforeseen side-effects when used within RF. It seems like RF itself is using the warnings module internally, but doesn't call this

packages/main/src/RPA/Robocorp/WorkItems.py
osrjv: Also in general I think the error can be a bit more verbose, and include the name of the env var in question

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -326,6 +333,20 @@ def get_input(self) -> str:
...
+    @property
+    def output_path(self):
+        if not self._output_path:
+            # This is usually set once per loaded input work item.
+            new_path = os.getenv("RPA_OUTPUT_WORKITEM_PATH")
+            if new_path:
+                self._output_path = resolve_path(new_path)
+                self._output_path.parent.mkdir(parents=True, exist_ok=True)
+            else:
+                deprecation("Work items save - Old path style usage detected")
```
osrjv, line 345: Same as input variable (docs, verbosity)

packages/main/tests/robot/test_robocorp_workitems.robot
```diff
@@ -16,10 +18,12 @@ Read input and write output
...
+    File should exist    ${temp_out}
 
 *** Keywords ***
 Load mock library
-    Copy file    ${RESOURCES}/items.json    ${RESOURCES}/temp_items.json
-    Set environment variable    RPA_WORKITEMS_PATH    \${RESOURCES}/temp_items.json
+    Copy file    ${RESOURCES}/items.json    ${temp_in}
+    Set environment variable    RPA_INPUT_WORKITEM_PATH    ${temp_in}
+    Set environment variable    RPA_OUTPUT_WORKITEM_PATH    ${temp_out}
+    remove file     ${temp_out}
```
osrjv, line 28: As a future reference, maybe prefer having clean-up as a teardown step:  `[Teardown]    Remove file    ${temp_out}`

packages/main/tests/robot/test_robocorp_workitems.robot
```diff
@@ -3,7 +3,9 @@ Suite setup    Load mock library
 Library        OperatingSystem
 
 *** Variables ***
-${RESOURCES}     ${CURDIR}/../resources
+${RESOURCES}    ${CURDIR}/../resources
+${temp_in}      ${RESOURCES}/temp_items.json
```
osrjv, line 7: Are these files gitignored?

packages/main/pyproject.toml
```diff
@@ -59,7 +59,7 @@ exchangelib = "^3.1.1"
 xlrd = "^2.0.1"
 xlwt = "^1.3.0"
 xlutils = "^2.0.0"
-openpyxl = "^3.0.3"
+openpyxl = "^3.0.9"
```
osrjv, line 62: I don't think we actually need to _depend_ on a newer version, but bumping the patch version doesn't hurt. The problem was the `poetry lock` generated lockfile that had a deleted version, and it could've been re-locked to fix.

Just being a bit pedantic here since these version ranges affect the wheels and the lockfiles don't. Probably no need to change it back though.

osrjv (commented):
Various small notes, looks good

packages/core/pyproject.toml
cmin764: We decided internally that's best to put all changes in one PR and then release updated dependencies to PyPI once the relevant CI for them passes here. (and once their changes are approved and stay backwards compatible)

packages/main/src/RPA/Robocorp/WorkItems.py
cmin764: Done, this is displayed like so now:
Screenshot 2021-09-27 at 12 19 50

----

The main advantage of it is that's displayed only once per line, so it comes handy when having a loop that would display N times the very same warning. (avoiding some logging spam)

packages/main/tests/robot/test_robocorp_workitems.robot
cmin764: Yes, there's a _resources/.gitignore_ with `temp_*.json` and makes more sense to move the output workitems in _results/_.

packages/core/src/RPA/core/logger.py
cmin764: Yes, definitely, good catch! Most probably would spam too much the dev.

I fixed it with a temporary switch to capturing logging only when using this utility. I've seen the idiom in some [examples](https://www.programcreek.com/python/example/107920/logging.captureWarnings).

packages/main/pyproject.toml
```diff
@@ -36,7 +36,8 @@ packages = [
 python = "^3.6.2"
 docutils = "*"
 dataclasses = { version = "^0.7", python=">=3.6,<3.7" }
-rpaframework-core = "^6.4.0"
+# rpaframework-core = "^6.5.0"
+rpaframework-core = { path = "../core", develop = true}
```
cmin764, line 40: Temporary change until `6.5.0` of **rpaframework-core** will be available, so the CI passes entirely.

packages/core/src/RPA/core/logger.py
```diff
@@ -11,6 +12,15 @@
 from RPA.core.helpers import required_param
 
 
+def deprecation(message):
+    """Emits once a deprecation warning log given the provided `message`."""
+    logging.captureWarnings(True)
+    try:
+        warnings.warn(message, FutureWarning, stacklevel=2)
+    finally:
+        logging.captureWarnings(True)
```
osrjv, line 21: Should this be `False`?

cmin764: Yes, dumb mistake :) I made.
Updated, also in the latest `6.5.1` hotfix of **rpaframework-core**.

docs/source/releasenotes.rst
```diff
@@ -4,6 +4,11 @@ Release notes
 Upcoming release
 ----------------
 
+11.3.0
```
cmin764, line 7: I'm not an expert on releases, so I'll need your guidance one more time @osrjv and @mikahanninen . Is it fine if I update the release notes along the changes in one go like so?

Post-merge, I would `% make publish` in _master_ **rpaframework** as `11.3.0`.

osrjv: Usually, changes are put under `Upcoming release` if there is no immediate plan to make one, but if you're releasing them soon anyways this will be fine. It's unavoidable that the release notes and PyPI release will be out of sync for a short while, but let's try to minimize that

osrjv: Assuming this is now tested and ready for release. If there's only a need to get a testable package for VSCode development, it's better to merge under `Upcoming` and distribute through devpi instead of PyPI.

PR #241, opened 2021-09-30
Helper keyword for iterating local work items

cmin764:
Add `@{results} =     For Each Input Work Item    <keyword>    <params>` keyword for consuming all work items in the local queue while applying a keyword over them.

Example
```robotframework
*** Settings ***
Documentation    Testing the helper keyword on consuming all work items in the queue.
Library          RPA.Robocorp.WorkItems
Library          OperatingSystem
Library          Collections


*** Variables ***
${idx}    ${1}


*** Keywords ***
Add Work Item
    [Arguments]  ${counter}

    # Get files from the current input work item.
    ${path} =        Get Work Item File    orders.txt    path=${OUTPUT_DIR}${/}orders${counter}.txt
    Log To Console    Getting: ${path}
    ${content} =     Get File    ${path}

    # Save a new file with the same content into one adjacent output work item.
    Create Output Work Item
    Set work item variables    name=CustomName${counter}   id=${counter}
    Add Work Item File    devdata/orders${counter}.txt
    Save Work Item
    [Return]    ${content}

Add Work Item Wrapper
    ${ret} =    Add Work Item    ${idx}
    Set Global Variable    ${idx}    ${idx + 1}

    # Any of these would fail if uncommented
    # Get Input Work Item
    # @{results} =     For Each Input Work Item    Add Work Item Wrapper
    
    [Return]    ${ret}


*** Tasks ***
Consume queue
    @{results} =     For Each Input Work Item    Add Work Item Wrapper
    Log List    ${results}
```

For the whole robot content including the prebuilt wheel with the changes in this PR see the attached archive. (make sure you rebuild your `rcc` env by removing the old one)

Latest robot (**4 Oct**): [work-items-matti.zip](https://github.com/robocorp/rpaframework/files/7278234/work-items-matti.zip)

_env.json_
```JSON
{
    "RPA_WORKITEMS_ADAPTER": "RPA.Robocorp.WorkItems.FileAdapter",
    "RPA_INPUT_WORKITEM_PATH": "%{ROBOT_ROOT}/devdata/items-in.json",
    "RPA_OUTPUT_WORKITEM_PATH": "%{ROBOT_ROOT}/output/items-out.json"
}
```

Look into the attached archive for the rest of the required files.

If it works, you should see some console output like so:
```console
Consume queue                                                         Getting: /Users/cmin/Robots/work-items-matti/output/orders1.txt
Getting: /Users/cmin/Robots/work-items-matti/output/orders2.txt
Getting: /Users/cmin/Robots/work-items-matti/output/orders3.txt
| PASS |
```

### Notes
- [Bugs](https://linear.app/robocorp/issue/LIB-12/fix-looping-of-work-items-bug) fixed
- While iterating work items already you can't (inside the iteration):
    - Iterate work items with the helper
    - Pull new input work items
    - Do it in the cloud (for now)

### ToDo
- [x] Confirmation from @janipalsamaki that this keyword works as expected
- [x] Unit and robot tests (adding/removing files under input/output work items, the helper keyword)
- [x] Review, refactor/accept,  release notes, squash-merge

### DIFF
--- a/packages/main/src/RPA/Robocorp/WorkItems.py
+++ b/packages/main/src/RPA/Robocorp/WorkItems.py
@@ -62,7 +62,7 @@ class BaseAdapter(ABC):
         raise NotImplementedError
 
     @abstractmethod
-    def add_file(self, item_id: str, name: str, original_name: str, content: bytes):
+    def add_file(self, item_id: str, name: str, *, original_name: str, content: bytes):
         """Attach file to work item."""
         raise NotImplementedError
 
@@ -175,7 +175,9 @@ class RobocorpAdapter(BaseAdapter):
 
         return response.content
 
-    def add_file(self, item_id: str, name: str, _: str, content: bytes):
+    def add_file(self, item_id: str, name: str, *, original_name: str, content: bytes):
+        # Note that here the `original_name` is useless thus not used.
+        del original_name
         # Robocorp API returns pre-signed POST details for S3 upload
         url = self.workitem_url(item_id, "files")
         info = {"fileName": str(name), "fileSize": len(content)}
@@ -415,7 +417,7 @@ class FileAdapter(BaseAdapter):
         with open(path, "rb") as infile:
             return infile.read()
 
-    def add_file(self, item_id: str, name: str, original_name: str, content: bytes):
+    def add_file(self, item_id: str, name: str, *, original_name: str, content: bytes):
         source, item = self._get_item(item_id)
         files = item.setdefault("files", {})
 
@@ -543,7 +545,9 @@ class WorkItem:
 
         for name, path in self._files_to_add.items():
             with open(path, "rb") as infile:
-                self.adapter.add_file(self.id, name, path.name, infile.read())
+                self.adapter.add_file(
+                    self.id, name, original_name=path.name, content=infile.read()
+                )
 
         # Empty unsaved values
         self._payload = self._payload_cache
--- a/packages/main/src/RPA/Robocorp/WorkItems.py
+++ b/packages/main/src/RPA/Robocorp/WorkItems.py
@@ -6,6 +6,7 @@ import os
 from abc import ABC, abstractmethod
 from pathlib import Path
 from shutil import copy2
+from threading import Event
 from typing import Type, Any, Optional, Union, Dict, List, Tuple
 
 import requests
@@ -786,6 +787,8 @@ class WorkItems:
         self._adapter_class = self._load_adapter(default_adapter)
         self._adapter: Optional[BaseAdapter] = None
 
+        self._under_iteration = Event()
+
     @property
     def adapter(self):
         if self._adapter is None:
@@ -864,7 +867,7 @@ class WorkItems:
         self.current = item
 
     @keyword
-    def get_input_work_item(self):
+    def get_input_work_item(self, _internal_call: bool = False):
         """Load the next work item from the input queue,
         and set it as the active work item.
 
@@ -875,8 +878,10 @@ class WorkItems:
         **NOTE**: Currently only one input work item per execution is supported
                   by Control Room.
         """
-        item_id = self.adapter.get_input()
+        if not _internal_call:
+            self._check_iteration_set("get input work item")
 
+        item_id = self.adapter.get_input()
         item = WorkItem(item_id=item_id, parent_id=None, adapter=self.adapter)
         item.load()
 
@@ -1258,6 +1263,10 @@ class WorkItems:
         logging.info("Removed %d file(s)", len(names))
         return names
 
+    def _check_iteration_set(self, action: str) -> None:
+        if self._under_iteration.is_set():
+            raise RuntimeError(f"Can't {action} while iterating input work items")
+
     @keyword
     def for_each_input_work_item(self, keyword, *args, **kwargs):
         """Run a keyword for each work item in the input queue.
@@ -1266,13 +1275,19 @@ class WorkItems:
         """
 
         outputs = []
-        while True:
-            output = BuiltIn().run_keyword(keyword, *args, **kwargs)
-            outputs.append(output)
+        self._check_iteration_set("iterate input work items")
+        self._under_iteration.set()
 
-            try:
-                self.get_input_work_item()
-            except EmptyQueue:
-                break
+        try:
+            while True:
+                output = BuiltIn().run_keyword(keyword, *args, **kwargs)
+                outputs.append(output)
+
+                try:
+                    self.get_input_work_item(_internal_call=True)
+                except EmptyQueue:
+                    break
+        finally:
+            self._under_iteration.clear()
 
         return outputs
--- a/packages/main/src/RPA/Robocorp/WorkItems.py
+++ b/packages/main/src/RPA/Robocorp/WorkItems.py
@@ -380,12 +380,7 @@ class FileAdapter(BaseAdapter):
         _, item = self._get_item(item_id)
         return item.get("payload", {})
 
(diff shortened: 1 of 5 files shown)


mikahanninen (commented):
interesting keyword, need to test myself before full approval

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -1255,3 +1271,32 @@ def remove_work_item_files(self, pattern, missing_ok=True):
...
+        Returns a list of results.
+        """
+
+        outputs = []
+        self.raise_under_iteration("iterate input work items")
+        self._under_iteration.set()
+
+        try:
+            while True:
+                output = BuiltIn().run_keyword(keyword, *args, **kwargs)
```
mikahanninen, line 1292: I guess that this row will fail on pure Python run

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -1255,3 +1271,32 @@ def remove_work_item_files(self, pattern, missing_ok=True):
...
         return names
+
+    def raise_under_iteration(self, action: str) -> None:
+        if self._under_iteration.is_set():
+            raise RuntimeError(f"Can't {action} while iterating input work items")
+
+    @keyword
+    def for_each_input_work_item(self, keyword, *args, **kwargs):
+        """Run a keyword for each work item in the input queue.
+
```
mikahanninen, line 1282: Add parameter descriptions and example how the keyword is used

fabioz (commented):
Hi @cmin764 I just did a quick review and ended up finding some nitpicks on general python usage, not real bugs... feel free to ignore if you don't think it's pertinent.

Some unrelated things which I noticed (which seem more related to the library, so, just raising the discussion and may not be pertinent to the actual pull request):

- Shouldn't there be Robot Framework tests accompanying the library? (it'd be nice to see in tests how the library is actually used)
- Is the library meant to be used without Robot Framework or only with Python? (some parts of the code seem to work only with Robot Framework...)

janipalsamaki:
I'm failing to figure out how to get access to the current work item (and its payload) in the callback. Any examples? :D

packages/main/src/RPA/Robocorp/WorkItems.py
cmin764: I added now the possibility of providing a Py function instead of just a RF keyword. Is that what we'd expect?

cmin764:
> * Framework tests accompanying the library? (it'd be nice to see in tests how the library is actually used)
> * Is the library meant to be used without Robot Framework or only with Python? (some parts of the code seem to work only with Robot Framework...)

Thanks for your attention to details.
- The library should work both under Py and RF, will take care of that.
- Yes, definitely I'll add both unit and robot tests. Just wanted the reviewers to have a glance at the logic first (thus marking the PR as draft) ahead of tests since is much faster to change code at this stage.

Will take care of your notes in this PR. (or another one with a Linear placeholder)

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -327,19 +336,21 @@ def _get_item(self, item_id: str) -> Tuple[str, Dict[str, Any]]:
...
+        if idx < (len(self.inputs) + len(self.outputs)):
             return "output", self.outputs[idx - len(self.inputs)]
-        else:
-            raise ValueError(f"Unknown work item ID: {item_id}")
+
+        raise ValueError(f"Unknown work item ID: {item_id}")
 
     def get_input(self) -> str:
+        idx = self.index
```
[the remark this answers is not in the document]
cmin764: Yes, that's an odd piece of code, refactored.

cmin764:
> I'm failing to figure out how to get access to the current work item (and its payload) in the callback. Any examples? :D

There's one example right in the PR [description](https://github.com/robocorp/rpaframework/pull/241#issue-746944016) -> check this part:
```robotframework
${path} =        Get Work Item File    orders.txt    path=${OUTPUT_DIR}${/}orders${counter}.txt
```
and here's another example:
```robotframework
Log Payload
    ${payload} =     Get Work Item Payload
    Log To Console    ${payload}
    ${len} =     Get Length    ${payload}
    [Return]    ${len}

Log Payloads
    @{results} =     For Each Input Work Item    Log Payload
    Log   Items keys length: @{results}
```
where the helper automatically loads the next work item in the queue and afterwards you can load its payload in the mapping keyword, like I'm doing above. This example can be found in the keyword docstring too now.

----
Tested with this wheel: [rpaframework-11.3.0-py3-none-any.whl.zip](https://github.com/robocorp/rpaframework/files/7278232/rpaframework-11.3.0-py3-none-any.whl.zip)
 
And here's the entire robot I used (including examples above and more): [work-items-matti.zip](https://github.com/robocorp/rpaframework/files/7278234/work-items-matti.zip)

cmin764:
### Caveats

- Using the `For Each Input Work Item` twice will re-pick the last item in the queue for executing the passed keyword over (and will happen the same each time you call it) -> might be solved in a polishing PR somewhere next week if necessary
- Providing explicitly `RPA_INPUT/OUTPUT...` env vars in the _env.json_ file will override the settings in the VSCode extension, thus not working as expected when picking up the next input work item for the robot run -> will be discussed in the extension call and fixed in the polishing PR

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -359,20 +366,27 @@ def output_path(self):
...
         return self._output_path
 
-    def create_output(self, parent_id: str, payload: Optional[JSONType] = None) -> str:
-        del parent_id
-
-        item: Dict[str, Any] = {"payload": payload, "files": {}}
-        self.outputs.append(item)
-
-        logging.debug("Payload: %s", json_dumps(payload, indent=4))
+    def _save_to_disk(self, source: str) -> None:
```
mmokko, line 369: Even easier to read code would be to have _save_input_to_disk / _save_output_to_disk and the do the file operation on the _save_to_disk function, would save a condition check, but that is matter of opinion of course :), anyway better this way than before. Just wanted to give any feedback :).

mmokko (approved):

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -61,7 +62,7 @@ def get_file(self, item_id: str, name: str) -> bytes:
         raise NotImplementedError
 
     @abstractmethod
-    def add_file(self, item_id: str, name: str, content: bytes):
+    def add_file(self, item_id: str, name: str, *, original_name: str, content: bytes):
```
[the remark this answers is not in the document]
cmin764: Every argument after `*` is necessary to be passed as a keyword argument when the function is called, without providing a default value for them (as in traditional keyword arguments). I added this here just to make sure the dev doesn't get lost in those 4 arguments the method receives.

packages/main/src/RPA/Robocorp/WorkItems.py
cmin764: Yes, I like this separation you suggest! Too bad I don't see how can I simplify the code within the current infrastructure. The culprit starts from this:
```python
    def _get_item(self, item_id: str) -> Tuple[str, Dict[str, Any]]:
        idx = int(item_id)
        if idx < len(self.inputs):
            return "input", self.inputs[idx]

        if idx < (len(self.inputs) + len(self.outputs)):
            return "output", self.outputs[idx - len(self.inputs)]

        raise ValueError(f"Unknown work item ID: {item_id}")
```
So solving that first might lead to a better codebase allowing this proposed refactor to happen.

PR #256, opened 2021-10-20
Provide exception types on input item release

cmin764:
With this an exception type (and code/message) can be provided with the usage of `Release Input Work Item` keyword during `FAILED` states only.

### ToDo
- [x] Params docstring & linting
- [x] Fix logging messages during releasing/reserving items
- [x] Unit & robot tests
- [x] Review, refactor/accept, release notes

### DIFF
--- a/packages/main/src/RPA/Robocorp/WorkItems.py
+++ b/packages/main/src/RPA/Robocorp/WorkItems.py
@@ -1474,7 +1474,7 @@ class WorkItems:
             Login into portal
                 ${user} =     Get Work Item Variable    user
                 Log     Logging in ${user}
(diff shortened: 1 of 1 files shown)


osrjv:
I haven't followed the conversation and I don't know the spec/reasoning behind this, but I guess my comments are mostly related to documentation.

It looks like the error code and message fields are both freeform strings, used for... whatever the user wants to I guess?
However, I don't exactly understand what it means to have a `FAILED` work item that doesn't have an `err_type`. The keyword documentation is phrased as `no exception is sent to Control Room`, but I assume some sort of error is. A less intelligent person (me) might ask why one user-facing enum isn't enough.

Probably quite self-evident if/when I have these new CR features in front of me, but the keyword documentation could try to explain a bit.

PR #277, opened 2021-11-03
`Parse Work Item From Email` keyword for dict body

cmin764:
Adds keyword for parsing e-mail JSON body when triggering processes by e-mail sending.

### ToDo

- [x] Keyword doesn't require to take raw e-mail content as argument. (can get that directly from the payload variable)
- [x] Parses the mail content and returns the body as dictionary. (data validation here, in case the sent body is not in JSON format)
- [x] Tests with different inputs, including attachments. (so the logic won't break, even if we don't support attachments at this stage -- just ignoring them)
- [x] Unit & robot tests
- [x] Review, refactor/accept, release note


packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -1579,3 +1581,27 @@ def get_current_work_item(self) -> WorkItem:
             Set Current Work Item    ${input}
         """
         return self.current
+
+    @keyword
+    def parse_work_item_from_email(self) -> dict:
+        """Parse and return a dictionary from the sent mail trigger.
```
mikahanninen, line 1587: ... dictionary from the input work item of a process started by email trigger

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -1579,3 +1581,27 @@ def get_current_work_item(self) -> WorkItem:
...
+
+        Since a process can be started in Control Room by sending an e-mail, a body
+        in JSON format can be sent as well and this gets attached to the input work
+        item with the `rawEmail` payload variable. This keyword parses the content of
+        it and returns the original dictionary that was sent through the mail.
```
mikahanninen, line 1592: returns the dictionary transformation of the original email.

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -1579,3 +1581,27 @@ def get_current_work_item(self) -> WorkItem:
...
+
+        Example:
```
mikahanninen, line 1595: this example is a bit odd to me

- why there is `Save Work Item`  ? 
- it is a bit unclear where `action` comes from, I suppose from email body. Maybe add multiline comment to show what email body actually looks like in this case ?

mikahanninen: or just 
# email body: {"action": "update"}
maybe ?

cmin764: Simplified it now:
- changed `action` -> `message`
- removed the save since it wasn't necessary
- added e-mail body example

mikahanninen (approved):
good to go after testing

cmin764:
Tested in CI with [robot](https://github.com/cmin764/robots/blob/main/work-items-loop/tasks.robot#L46) and then triggered e-mail with JSON content:
Screenshot 2021-11-03 at 17 27 22
and got log:
Screenshot 2021-11-03 at 17 27 29

----

Then ran the same with attachment -> run is not triggered by the cloud (maybe because attachments aren't supported or mine was too big (a zip of ~200kb))

Test runs 1 & 3 - without attachments (only 2nd run with attachment)
Screenshot 2021-11-03 at 17 31 38

as you can see only the first and 3rd run were triggered:
Screenshot 2021-11-03 at 17 35 27

PR #284, opened 2021-11-05
Skip unexpected errors from retrying during work items API calls

cmin764:
If `ERR_UNEXPECTED` just changed in the cloud into `UNEXPECTED_ERROR` (as **error.code**), then this needs to be reflected in the library as well in order to skip from retrying any unexpected error as well.

image

- Slack [thread](https://robocorptechnologies.slack.com/archives/C013FUJQVL4/p1636122677117500)


cmin764:
Actually we need to do the [opposite](https://github.com/robocorp/rpaframework/issues/298).

PR #285, opened 2021-11-05
Support JSON/YAML/Text input e-mail body content as work item

cmin764:
### ToDo

- [ ] ~Move keyword into `RPA.Email` package. (make it take the raw e-mail content as argument)~
- [x] Eliminate the helper keyword and do the parsing automatically when the work item is loaded
- [x] Configureable the mail content transformation key: `{"rawEmail": "parsedEmail"}` (and a falsy value for not making the conversion at all)
- [ ] Run `poetry update` on the upper meta-package
- [x] Review, refactor/accept


cmin764:
Tested in the cloud with task:

```robotframework
Get payload given e-mail process triggering
    ${mail} =    Get Work Item Variable    parsedEmail
    Log    ${mail}
    Set Work Item Variables    &{mail}[Body]
    Save Work Item
    ${message} =     Get Work Item Variable     message
    Should Be Equal     ${message}      from email
```

and mail triggers like:

```json
{
    "message": "from email",
    "extra": "value"
}
```

```yaml
message: from email
extra:
  my: value
  is: 1
```

and it works as expected.
Screenshot 2021-11-16 at 13 30 16
Screenshot 2021-11-16 at 13 30 50
Screenshot 2021-11-16 at 13 32 59

docs/source/releasenotes.rst
```diff
@@ -4,6 +4,11 @@ Release notes
 `Upcoming release <https://github.com/robocorp/rpaframework/projects/3#column-16713994>`_
 +++++++++++++++++
 
+- Library **RPA.Robocorp.WorkItems**:
+  - Removed Keyword ``Parse Work Item From Email``
```
cmin764, line 8: This will be a breaking change, therefore bumping **rpaframework** to `12.x.x`.

Issue #283, opened 2021-11-05
PDF library outputs duplicates to XML in notebook


orlof:
Paste following code into Lab cell and run it twice with SHIFT+ENTER
First run produces correct XML, but second run produces the same XML twice into the voice.xml - even if you delete voice.xml between runs. 
```
*** Settings ***
Library  RPA.FileSystem
Library  RPA.PDF

*** Tasks ***
Testing
    Get text from PDF  voice.pdf
    ${xml}  Dump PDF as XML  voice.pdf
    create file  voice.xml  ${xml}  overwrite=TRUE
```

cmin764:
Might be related to that context ensuring of a PDF file/content which might interfere with the currently running kernel compared to the Python process that exits in the `rcc run` and VSCode cases. Similar to the empty XML [issue](https://github.com/robocorp/rpaframework/issues/243#issue-1015068830) case (when not reading some text first).

cmin764:
This is solved now with the upcoming [fix](https://github.com/robocorp/rpaframework/pull/288) ([loom](https://www.loom.com/share/18ec3266048d42daa96a2101944fec0f) recording)

PR #288, opened 2021-11-10
Fix PDF document active content and XML dumping

cmin764:
Fixes issues when switching to existing/new PDF documents and XML parsing document content. (and some closely related refactor to make the logic understandable)

### ToDo

- [x] XML & text conversion unit and robot test
- [x] Add test for encrypted PDF case (check `reader.isEncrypted` attr)
- [x] Test robot in CI and locally on Mac & Windows
- [x] Check robot in lab with XML dump -> see if output gets duplicated on [consecutive](https://github.com/robocorp/rpaframework/issues/283) runs
- [x] Optional improvement: when caching already opened file objects, save `Document`s instead, so we preserve its entire state rather than `reinitialize`-ing it partially (some extra work required)
- [x] Make robot tests run in CI too (within _Makefile_)
- [ ] Publish **rpaframework-pdf** `0.10.0` (will be picked-up automatically by new env installs of the _main_ package)

### DIFF
--- a/packages/pdf/src/RPA/PDF/__init__.py
+++ b/packages/pdf/src/RPA/PDF/__init__.py
@@ -61,7 +61,7 @@ class PDF(DynamicCore):
 
     def __init__(self):
         self.logger = logging.getLogger(__name__)
-        self.fileobjects = {}
+        self.documents = {}
         self.active_pdf_document = None
         self.convert_settings = {}
 
@@ -76,4 +76,4 @@ class PDF(DynamicCore):
         listener = RobotLogListener()
         listener.register_protected_keywords(["RPA.PDF.decrypt"])
 
-        logging.getLogger("pdfminer").setLevel(logging.INFO)
+        logging.getLogger("pdfminer").setLevel(logging.WARNING)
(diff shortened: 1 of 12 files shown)


packages/pdf/src/RPA/PDF/__init__.py
```diff
@@ -76,4 +76,4 @@ def __init__(self):
         listener = RobotLogListener()
         listener.register_protected_keywords(["RPA.PDF.decrypt"])
 
-        logging.getLogger("pdfminer").setLevel(logging.INFO)
+        logging.getLogger("pdfminer").setLevel(logging.WARNING)
```
cmin764, line 79: There was too much **pdfminer** related output in the logs.

cmin764:
Was able to run the [robot](https://github.com/cmin764/robots/blob/main/pdf-parse/tasks.robot#L49) without issues on:

- Mac/Windows with VSCode (and `% rcc run`)
- Lab (no junk XML on subsequent individual cell run)
- Production Control Room

Screenshot 2021-11-11 at 16 14 22

PR #300, opened 2021-11-19
PDF improvements on `Find Text` (breaking changes)

cmin764:
### ToDo
- [x] Multiple anchor points supported
- [x] Direction traversal for 0, 1, 2, ..., N closest neighbour elements starting from each anchor
- [x] `Find Text` returns `Match` objects with `.anchor` and `.neighbours`
- [x] Text locator as regular expression
- [x] Optimized fast search based on the page we're looking into
- [x] Reduce search complexity (if possible/reasonable, from N^2 to NlogN -> not feasible since each sorting depends on the selected anchor)
- [x] Unit & robot tests
- [x] Improved debug logs and docstrings
- [x] Bump to version `1.26.11` (to avoid `1.0.0`)
- [x] Windows & CI test
- [x] Review, refactor/accept


packages/pdf/src/RPA/PDF/keywords/finder.py
```diff
@@ -1,49 +1,94 @@
...
+
+        # Text locator might lead to multiple valid found anchors.
+        self._anchors: List[Union[TargetObject, TextBox]] = []
+        # The others usually have just one. (if multiple are found, set to it the
+        #   first one)
         self.anchor_element = None
 
+    def _get_candidate_search_function(
+        self, direction: str, regexp: Optional[re.Pattern], strict: bool
+    ) -> Callable[[TextBox], bool]:
```
cmin764, line 54: Fix the `Callable` args.

cmin764:
I'll be taking care of the `re.Pattern` missing in Py 3.6 thingie.

packages/pdf/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "rpaframework-pdf"
-version = "0.10.0"
+version = "1.26.11"
```
cmin764, line 3: Because of major version bump to `1.0.0` and because people don't trust "beta" versions like these :D.

PR #313, opened 2021-12-07
Add missing fonts for unicode PDFs

cmin764:
When rendering HTML with strong/bold, emphasis/italic tags, certain Inter font styles like `B`, `I` and `BI` are requested, so we made sure we added them in the PDF library assets.


cmin764:
Tested locally on both Mac and Windows and with uploaded robot into production CI and it works flawlessly now:

Screenshot 2021-12-07 at 11 33 29
Screenshot 2021-12-07 at 12 45 07
Screenshot 2021-12-07 at 12 45 23

PR #339, opened 2021-12-15
Fix bug when adding a watermark into the same PDF

cmin764:
Uses an auxiliary buffer for reading/writing into the same PDF file that acts as both the source and the destination.

```robotframework
Add watermark into PDF
    ${in_out_pdf} =     Set Variable    ${OUTPUT_DIR}${/}receipt.pdf
    Copy File    devdata${/}receipt.pdf    ${in_out_pdf}
    ${screenshot} =    Set Variable    devdata${/}robot.png
    Open Pdf    ${in_out_pdf}
    Add Watermark Image to PDF    ${screenshot}    ${in_out_pdf}
```


cmin764:
Tested on Mac, Windows and Control Room with [robot](https://github.com/cmin764/robots/blob/main/pdf-parse/tasks.robot#L185).

PR #340, opened 2021-12-15
Add migration guide on >=11 library update

cmin764:


docs/source/releasenotes.rst
```diff
@@ -227,6 +227,17 @@ Releases on 01 Dec 2021
 11.0.0
 ------
 
+- Migration guide: Given this major upgrade, the ``Load Work Item ...`` keywords got
```
cmin764, line 230: This is how it's rendered:

PR #351, opened 2021-12-30
HTTP.py defined and described return values

TimothyLopas:
Defined the -> return type and described the :return: for the HTTP.py file


mikahanninen (changes requested):
I think these changes won't work because library doc string formatting is not REST

`ROBOT_LIBRARY_DOC_FORMAT = "ROBOT"`

cmin764 (commented):
LGTM!

And as @mikahanninen said, I guess `ROBOT_LIBRARY_DOC_FORMAT = "REST"` should be placed instead in the class. Look for more info [here](https://robotframework.org/robotframework/4.0/RobotFrameworkUserGuide.html#documentation-format).

TimothyLopas:
Based on the documentation in Robot Framework it appears that no additional changes are necessary to the doc strings after the `ROBOT_LIBRARY_DOC_FORMAT`  was updated to `"reST".`

cmin764 (approved):
Looks good now, I think you can merge without an _Upcoming_ release note since is just docstring docs reformatting.

PR #373, opened 2022-02-01
Support for locators path configuration

osrjv:
Previously `RPA.Browser.Selenium` supported configuring the path to the locators database, but that support was missing from `RPA.Desktop`.

This PR adds the following:
- Support for configuring the database path used in the syntax parser
- Exposing the path as a `RPA.Desktop` import argument
- Support for the environment variable `RC_LOCATORS_DATABASE`, which allows configuring it for all libraries

Do before merge:
- Actually test the features
- Decide on final environment variable name


cmin764 (approved):
LGTM! The env var looks fine.

mikahanninen:
Seems all right. Still on a fence if RPA_LOCATORS_DATABASE would be a better env name because a similar prefix is used by Vault and Workitems.

Have to find time for testing this.

cmin764:
- I went with the more standard `RPA_LOCATORS_DATABASE` env var. (similar to what we find in `Vault` and `WorkItems`)
- As `RPA.Desktop` will depend on this future `core` release, I created a new [PR](https://github.com/robocorp/rpaframework/pull/592) tackling this.

⚠️ Question: Is there any sane case (beside unit-testing) where we actually pass as the `locators_path` an open stream object instead of a file path (str)? -- Asking because we close the file descriptor now when existing the `with` context. @mikahanninen @osrjv

cmin764:
Ran a manual test for a robot with a custom path given the _locators.json_ file and it worked ok:
```console
RPA_LOCATORS_DATABASE=$HOME/locators.json python -m robot -L TRACE -d ./output --task "My Process" tasks.robot
```

PR #374, opened 2022-02-01
Windows library update rebased

mmokko:


packages/core/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "rpaframework-core"
-version = "6.5.1"
+version = "6.5.2"
```
cmin764, line 3: Since we're upgrading the core version here, after merge and release, we should upgrade to `rpaframework-core = "^6.5.2"` dependency in **rpaframework-windows** as well, right?

cmin764 (approved):
LGTM, although still fails on linting with `poetry run black --check src/RPA tests/` (run `poetry run black src/RPA tests/` to auto-format inside the affected packages)

Would love a description either in the PR or the linked issue (if any created already) of what this PR does and how `WindowsLocator` will be used. (I understood so far that now we're able to get icons too, but it seems to be more here with the **core** package upgrade)

cmin764:
Now [fails](https://github.com/robocorp/rpaframework/runs/5018325708?check_suite_focus=true) with: `Version 3.6 with arch x64 not found`

We started to remove Py3.6 support from the GitHub flows.

PR #375, opened 2022-02-01
Fix window/anchor context and new `Close Window` keyword

cmin764:
- Fixes a bug (COMError when searching element starting from closed window as root) reproduced with [this](https://github.com/cmin764/robots/blob/main/desktop-auto/tasks.robot#L125-L129) robot.
- Adds a new `Close Window` keyword which properly closes any matched open window.
- Fixes `timeout` setting/unsetting during keywords usage that call other keywords. (decorates keywords with `@with_timeout`)
- Necessary refactors to make code readable and sane + consistency.
- Added basic sibling check. (read FIXME)

### ToDo
- [x] Add logic for the new `Close Window` keyword.
- [x] Clean anchor context if this window gets closed too during `Close Window`.
- [x] Fix breaking `Print Tree`. (crashes on _master_ as well)
- [x] Add sweet unit & robot tests.
- [x] Review, refactor, accept.


cmin764:
Tested the attached wheel successfully with this RF code:
```robotframework
Control window after closing linked root
    [Setup]    Close all Calculators and open Notepad
    ${window} =     Windows.Control Window   subname:Notepad   timeout=1
    Log    Controlling Notepad window: ${window}

    Kill app by name    Notepad

    Windows.Windows Run   Calc
    # "COMError: (-2147220991, 'An event was unable to invoke any of the subscribers', (None, None, None, 0, None))"
    # Happens due to `str(self.ctx.window)` over a window that doesn't exist anymore.
    # Solution: cleanup context properly (through `Close Window` keyword) and tackle
    #  internally this edge case.
    ${window} =     Windows.Control Window   subname:Calc   timeout=1
    Log    Controlling Calculator window: ${window}

    [Teardown]    Windows.Close Current Window
```

[rpaframework_windows-2.1.0-py3-none-any.whl.zip](https://github.com/robocorp/rpaframework/files/7988721/rpaframework_windows-2.1.0-py3-none-any.whl.zip)

packages/windows/src/RPA/Windows/keywords/locators.py
```diff
@@ -70,6 +70,14 @@ def __init__(self, item, locator):
         self.xcenter = rect.xcenter() if rect else -1
         self.ycenter = rect.ycenter() if rect else -1
 
+    def is_sibling(self, win_elem: "WindowsElement") -> bool:
+        """Returns `True` if the provided window element is a sibling."""
+        # FIXME(cmin764): Is this enough to recognize the provided element as a
```
cmin764, line 75: @mikahanninen Do you think this approach is satisfying for now?

cmin764:
Tested with this [robot](https://github.com/cmin764/robots/blob/main/desktop-auto/tasks.robot#L118-L140) tasks:
```robotframework
*** Settings ***
Library    RPA.Windows    WITH NAME    Windows


*** Keywords ***
Open a single Notepad
    Windows.Set Global Timeout    1
    ${closed} =    Windows.Close Window    subname:Notepad control:WindowControl  # in development keyword
    Log    Closed Notepads: ${closed}
    Windows.Windows Run   Notepad


*** Tasks ***
Control window after closing linked root
    [Setup]    Open a single Notepad
    ${window} =     Windows.Control Window   subname:Notepad control:WindowControl
    Log    Controlling Notepad window: ${window}

    Kill app by name    Notepad

    Windows.Windows Run   Calc
    # "COMError: (-2147220991, 'An event was unable to invoke any of the subscribers', (None, None, None, 0, None))"
    # Happens due to `str(self.ctx.window)` over a window that doesn't exist anymore.
    # Solution: cleanup context properly (through `Close Window` keyword) and tackle
    #  internally this closed `window` edge case.
    ${window} =     Windows.Control Window   subname:Calc
    Log    Controlling Calculator window: ${window}

    [Teardown]    Windows.Close Current Window  # closes Calculator (last active window)

Control anchor cleanup
    Windows.Windows Run   Calc
    ${win} =    Windows.Control Window   subname:Calc control:WindowControl    timeout=1
    Windows.Set Anchor    ${win}
    Windows.Close Window    subname:Calc control:WindowControl    timeout=1
```

packages/windows/src/RPA/Windows/keywords/action.py
```diff
@@ -123,19 +124,15 @@ def middle_click(
         return self._mouse_click(locator, "MiddleClick", wait_time, timeout)
 
     def _mouse_click(self, element, click_type, wait_time, timeout):
-        if timeout:
-            auto.SetGlobalSearchTimeout(timeout)
         click_wait_time = wait_time or self.ctx.wait_time
-        try:
+        with self.set_timeout(timeout):
```
mmokko, line 128: Really nice use of context manager here and a text book place where to utilize it 👍 Makes code much more readable and you never forget to set the old value back.

packages/windows/src/RPA/Windows/keywords/context.py
```diff
@@ -33,5 +46,65 @@ def logger(self):
...
-    def window(self):
-        return self.ctx.window
+    def current_timeout(self) -> float:
+        # This value can change based on `auto.SetGlobalSearchTimeout(...)` calls.
+        return auto.uiautomation.TIME_OUT_SECOND
+
+    def _window_or_none(self, window) -> Optional["WindowsElement"]:  # noqa: F821
+        if window and window.item:
+            # FIXME(cmiN): Investigate why `.Exists()` is `False` with Desktop element.
+            #  (`auto.GetRootControl().Exists()` -> `False`)
```
mmokko, line 56: I guess the root control always return false as it doesn't have visible window. You can get windows from that by calling:  auto.GetRootControl().GetChildren(). Which then return list of auto.Control objects that has the Name and handle objects.

Another way of getting windows directly from the windows API would be:
win32gui.FindWindow(None, "Calculator")
Though I guess we don't want to have too many libraries to get the windows elements. Just makes things confusing.

It would also be nice to have the type declaration for the parameters. For example here I guess this is: window: auto.Control.

packages/windows/src/RPA/Windows/keywords/locators.py
```diff
@@ -70,6 +79,23 @@ def __init__(self, item, locator):
         self.xcenter = rect.xcenter() if rect else -1
         self.ycenter = rect.ycenter() if rect else -1
 
+    def is_sibling(self, win_elem: "WindowsElement") -> bool:
+        """Returns `True` if the provided window element is a sibling."""
+        locator: Optional[Locator] = win_elem.locator
+        while locator:
+            if isinstance(locator, WindowsElement):
+                locator = locator.locator
+        if not locator:
```
mmokko, line 88: Can you use here the rare while else syntax? :)
while locator:
    Do stuff
else:
    Nothing to do

packages/windows/src/RPA/Windows/keywords/window.py
```diff
@@ -153,17 +158,17 @@ def foreground_window(
...
-        if not self.ctx.window.item:
+        window = self.window
+        if window is None:
             raise WindowControlError("There is no active window")
-        utils.call_attribute_if_available(self.ctx.window.item, "SetFocus")
-        utils.call_attribute_if_available(self.ctx.window.item, "SetActive")
-        self.ctx.window.item.MoveCursorToMyCenter(simulateMove=self.ctx.simulate_move)
-        return self.ctx.window
+
+        utils.call_attribute_if_available(window.item, "SetFocus")
```
mmokko, line 165: Nice wrapper for safety 👍

mmokko (commented):
Great changes, really like to have better typing there, much easier to read the code.

I'll do some testing on windows and make another review once it's done.

packages/windows/src/RPA/Windows/__init__.py
```diff
@@ -466,13 +466,13 @@ def test_do_some_calculations():
     ROBOT_LIBRARY_SCOPE = "GLOBAL"
     ROBOT_LIBRARY_DOC_FORMAT = "REST"
 
-    def __init__(self, **kwargs):
+    def __init__(self, locators_path: Optional[str] = None):
```
cmin764, line 469: Is it better like this @osrjv / @mmokko ? (just a nitpick along the way)

packages/windows/src/RPA/Windows/keywords/action.py
cmin764: Thx! Also not to forget that the previous behaviour was bugged (so this is not just a structural refactor), as a keyword that was setting a specific timeout was calling another that was either setting/unsetting it again in some bad places and that leaved branches of logic unguarded by the correct desired timeout setting. (or guarded by the wrong timeout amount)

packages/windows/src/RPA/Windows/keywords/context.py
cmin764: Thx!
- Added "WindowsElement" as `window` typing. (since that's what we're receiving and returning)
- Given the access of Windows, I'm only interested into one thing: is the received control (usually app window) still existing on the screen or not? (and testing it with either of the bounding rectangle or children attribute produces the same COMError I was looking for)

```
>>> ca.BoundingRectangle
Rect(77,64,743,1142)[666x1078]
>>> ca.BoundingRectangle
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "Z:\Repos\rpaframework\packages\windows\.venv\lib\site-packages\uiautomation\uiautomation.py", line 5906, in BoundingRectangle
    rect = self.Element.CurrentBoundingRectangle
_ctypes.COMError: (-2147220991, 'An event was unable to invoke any of the subscribers', (None, None, None, 0, None))
>>> ca.GetChildren()
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "Z:\Repos\rpaframework\packages\windows\.venv\lib\site-packages\uiautomation\uiautomation.py", line 6341, in GetChildren
    child = self.GetFirstChildControl()
  File "Z:\Repos\rpaframework\packages\windows\.venv\lib\site-packages\uiautomation\uiautomation.py", line 6287, in GetFirstChildControl
    ele = _AutomationClient.instance().ViewWalker.GetFirstChildElement(self.Element)
_ctypes.COMError: (-2147220991, 'An event was unable to invoke any of the subscribers', (None, None, None, 0, None))
```

Now since Desktop makes sense of not having an active Window, I'll just remove the `.Exists()` block to make the test much more simpler. (by testing against the roots of the problem)

packages/windows/src/RPA/Windows/keywords/locators.py
cmin764: That was actually a bug where I missed this:
```python
            else:  # reached a string/`None`
                break
```
which created an infinite loop.

Now that I added this, there might be cases where the initial locator is a `WindowsElement` and the accessed inner locator is `None` so I still need to check both this case (and the one where the initial locator is `None`) within a single checkpoint as of:
```python
        if not locator:
            return True  # nothing to check, can be considered sibling
```

packages/windows/src/RPA/Windows/keywords/window.py
cmin764: Thx, although you might get tricked by the diff as it wasn't created/edited by me.
The refactor was: `self.ctx.window` -> `self.window`, so it gets through property wrapper's checks.

cmin764:
Added Mika again in this PR for post-merge visibility reasons only. (no need to review as he's busy)

packages/windows/src/RPA/Windows/keywords/context.py
```diff
@@ -33,5 +46,58 @@ def logger(self):
...
+    def set_timeout(self, timeout: Optional[float] = None):
+        """Context manager that sets a custom temporary `timeout`."""
+        try:
+            if timeout is not None:
+                auto.SetGlobalSearchTimeout(timeout)
+                self.logger.info("Locator timeout set to: %f", self.current_timeout)
+            yield
+        finally:
+            if timeout is not None:
+                auto.SetGlobalSearchTimeout(self.ctx.global_timeout)
```
osrjv, line 86: is there any situation where these context managers would be nested? just wondering if the timeouts should be a stack or something

cmin764:
Going to merge and release as the last two added [tests](https://github.com/robocorp/rpaframework/pull/375/files#diff-7133944bd147c0b270337b7431d2bd3b7f90e98c899ecb6a8a3b11eda53f7fe4R237-R261) are passing gracefully on Windows 11. (and most probably on 10 too -- @mikahanninen can check it afterwards)

Small downside of this PR: CI is failing on Linux/Mac due to **comtypes** `Syntax Error` issue that doesn't seem to be solved on these systems due to `pip install` instability. (our **comtypes-fork** doesn't get to replace the problematic **comtypes** properly -> Solution: wait for the official `1.1.11` upcoming release and upgrade to that while removing our **-fork**)

Any post-merge follow-up comment will be addressed in another PR.

packages/windows/tests/robot/test_windows.robot
```diff
@@ -211,3 +233,29 @@ Test get element
...
+    Log    Controlling Calculator window: ${window}
+
+    [Teardown]    Close Current Window  # closes Calculator (last active window)
+
+Tree printing and controlled anchor cleanup
+    Print Tree     #capture_image_folder=output${/}controls
+    Windows Run   Calc
+    ${win} =    Control Window   subname:Calc control:WindowControl    timeout=${TIMEOUT}
+    Set Anchor    ${win}
+    Close Window    subname:Calc control:WindowControl    timeout=${TIMEOUT}
```
osrjv, line 256: should you also check the return value in these tests?

packages/windows/src/RPA/Windows/keywords/window.py
```diff
@@ -367,14 +376,64 @@ def close_current_window(self) -> bool:
...
             return False
-        pid = self.ctx.window.item.ProcessId
-        name = self.ctx.window.item.Name
-        self.ctx.logger.info(
-            'Closing window with Name:"%s", ProcessId: %s' % (name, pid)
-        )
+
+        pid = window.item.ProcessId
+        self.logger.info("Closing window with name: %s (PID: %d)", window.name, pid)
         os.kill(pid, signal.SIGTERM)
```
osrjv, line 387: This is more me wondering about the Python API, but it's funky that we use kill with SIGTERM here since a) there is no SIGTERM signal on Windows and b) this library is Windows only. I know that os.kill straight up kills the process as a compatibility thing, but feels... odd.

packages/windows/src/RPA/Windows/keywords/window.py
```diff
@@ -367,14 +376,64 @@ def close_current_window(self) -> bool:
...
+                break
+        if not elements:
+            self.logger.info("Couldn't find any window with locator: %s", locator)
+            return 0
+
+        closed = 0
+        for element in elements:
+            self.logger.debug("Controlling and closing window: %s", element)
+            try:
+                self.control_window(element)
```
osrjv, line 435: what's the "controlled window" after these close calls? sorry if that's obvious from reading the library docs

packages/windows/src/RPA/Windows/keywords/window.py
```diff
@@ -367,14 +376,64 @@ def close_current_window(self) -> bool:
...
+        elements: Optional[List[WindowsElement]] = None
+        for loc in self._iter_locator(locator):
+            try:
+                elements = self.ctx.get_elements(loc, root_element=root_element)
+            except (ElementNotFound, LookupError):
+                pass
+            else:
+                break
+        if not elements:
+            self.logger.info("Couldn't find any window with locator: %s", locator)
```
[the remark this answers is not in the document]
cmin764: Good spot, thx! In order to be consistent (with `Control Window`), indeed we need to raise if no window is found at all, fixed! (and for the cleanup to work, you have to run `Close Window` with [Run Keyword And Ignore Error](https://robocorp.com/docs/libraries/built-in/builtin/keywords#run-keyword-and-ignore-error))

packages/windows/src/RPA/Windows/keywords/context.py
cmin764: Luckily, no, otherwise we should stack the timeouts.

For now the principle is simple: once a timeout is set for a keyword (and you're under the run of this keyword), no matter how many others keyword/methods we call, the original timeout remains set and no other timeout overrides the currently set one. (so the initial set timeout is inherited in all the other stacked calls)

packages/windows/src/RPA/Windows/keywords/window.py
cmin764: I guess you wanted to say there's no `SIGKILL` on Windows (while on *nix there are both):
```python
>>> import signal
>>> signal.SIGKILL  # happens on Windows only (exists on Mac/Linux)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'signal' has no attribute 'SIGKILL'
>>> signal.SIGTERM
<Signals.SIGTERM: 15>
```

To my understanding `os.kill` is more like a badly named function for interrupting a process with a mandatory signal parameter.
> The SIGTERM signal is a generic signal used to cause program termination. Unlike SIGKILL , this signal can be blocked, handled, and ignored. It is the normal way to politely ask a program to terminate. The shell command kill generates SIGTERM by default.

https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html#:~:text=The%20SIGTERM%20signal%20is%20a,kill%20generates%20SIGTERM%20by%20default.

packages/windows/src/RPA/Windows/keywords/window.py
cmin764: - Each time a window is controlled -> it becomes the active window (only one active/controlled window can exist at a time)
- Each active window gets closed -> the current `self.ctx.window_element` becomes None (after closing per se)

So no matter how many close calls (>=1) you have, in the end there's no current active window set. (as `Close Window` is more like a wrapper over the following: find me all the windows matching `locator`, then activate and close each of them, one by one)

packages/windows/src/RPA/Windows/keywords/window.py
osrjv: I was referring to the fact that Windows doesn't have POSIX style signals at all. It has events for interrupts and methods to terminate processes, but not signals. There are just various places where they're automatically handled for compatibility reasons. For os.kill [the documentation](https://docs.python.org/3/library/os.html#os.kill) says the following:

> Windows: The `signal.CTRL_C_EVENT` and `signal.CTRL_BREAK_EVENT` signals are special signals which can only be sent to console processes which share a common console window, e.g., some subprocesses. Any other value for sig will cause the process to be unconditionally killed by the TerminateProcess API, and the exit code will be set to sig. The Windows version of `kill()` additionally takes process handles to be killed.

Not that it really matters here. That function call does what we want, it's just a bit odd.

PR #382, opened 2022-02-04
Support for loading windows locators from DB

mmokko:


packages/windows/src/RPA/Windows/keywords/locators.py
```diff
@@ -277,6 +288,8 @@ def get_element(
             Set Value   ${element}  note to myself
         """
         # TODO. Add examples
+        if isinstance(locator, str):
+            locator = self._load_by_alias(locator)
```
cmin764, line 293: This is a good place, well done! (since this is the source of truth for getting elements)

packages/windows/src/RPA/Windows/keywords/context.py
```diff
@@ -25,8 +25,9 @@ class ActionNotPossible(ValueError):
 class LibraryContext:
     """Shared context for all keyword libraries."""
 
-    def __init__(self, ctx):
+    def __init__(self, ctx, *args, **kwargs):
         self.ctx = ctx
+        self._locators_path = kwargs.get("locators_path", None)
```
cmin764, line 30: Since this is used inside of `LocatorKeywords` only, I would place it there only. (in `LocatorKeywords` class -- then no need for args/kwargs init idiom)

packages/windows/src/RPA/Windows/__init__.py
```diff
@@ -466,7 +466,7 @@ def test_do_some_calculations():
     ROBOT_LIBRARY_SCOPE = "GLOBAL"
     ROBOT_LIBRARY_DOC_FORMAT = "REST"
 
-    def __init__(self):
+    def __init__(self, *args, **kwargs):
```
cmin764, line 469: Not sure I get where these args/kwargs are used. Can you point it out?

cmin764 (changes requested):
Looks good so far, a few edits (or clarifications) needed although.

packages/windows/src/RPA/Windows/keywords/context.py
mmokko: That's a good point.

packages/windows/src/RPA/Windows/__init__.py
mmokko: So when initializing the windows library you need a way to set the locators path. The way windows library is initialized in the test_windows.py file.

I noticed that the *args, **kwargs idiom was used in the selenium library so wanted to follow that.

packages/windows/src/RPA/Windows/keywords/locators.py
```diff
@@ -241,6 +242,17 @@ def _get_element_with_locator_part(
         new_element.robocorp_click_offset = offset
         return new_element
 
+    def _load_by_alias(self, criteria: str) -> str:
+        try:
+            locator = LocatorsDatabase.load_by_name(criteria, self._locators_path)
+            if isinstance(locator, WindowsLocator):
+                return locator.value
```
cmin764, line 249: So if the obtained locator through criteria isn't found (or anything that makes it to not be an instance of `WindowsLocator `), then the criteria itself is returned (as the original passed locator), right? (wanted to make sure I understand it)

mmokko: That's right. I know the naming is a bit confusing there as the str value of criteria can be a raw locator string as well. There is no way of knowing though if it's a criteria (key in locators.json) or a raw locator value (for example: name:'Number pad' > name:'Nine').

packages/windows/src/RPA/Windows/__init__.py
cmin764: Gotcha, now I saw the `kwargs.get("locators_path", None)` line, all good! Still, you can place a `locators_path: Optional[str] = None` in a new `LocatorKeywords.__init__` (and keeping the variable args/kwargs when passing from the Windows library to this class instantiation, like you already did). Don't forget to `super().__init__` the initializer from `LocatorKeywords` into `LibraryContext` too.

packages/windows/src/RPA/Windows/__init__.py
```diff
@@ -485,7 +485,7 @@ def __init__(self):
         libraries = [
             ActionKeywords(self),
             ElementKeywords(self),
-            LocatorKeywords(self),
+            LocatorKeywords(self, kwargs.get("locators_path", None)),
```
cmin764, line 488:
```suggestion
            LocatorKeywords(self, **kwargs),
```
can be more generic if you like it.

cmin764 (approved):
LGTM, good job! (added a tiny nitpick though)

packages/windows/src/RPA/Windows/keywords/locators.py
```diff
@@ -241,6 +246,17 @@ def _get_element_with_locator_part(
...
+        except ValueError:
+            # How to check if locator check should be done as inspector
```
osrjv, line 255: This could be handled more nicely with in the future if we harmonize the syntax parsing between `RPA.WIndows` and `RPA.Desktop`

packages/windows/src/RPA/Windows/__init__.py
osrjv, line 488: This is a matter of taste, but I'd almost have all of the library attributes stored in this container class, and accessing them through `self.ctx `in the keyword classes.

osrjv: Take a look at this pending PR for instance: https://github.com/robocorp/rpaframework/pull/373

cmin764: That's because you're accessing the locators from multiple classes, right? Here they're required in the `LocatorKeywords` only, but like you said, it's about a taste of opening them more to a global access or restricting to the place they're used only. (until further access extension required)

osrjv: True enough. Actually the only thing that's a bit of a downside in that diff is the use of **kwargs, which prevents IDEs and documentation from showing what the actual library init arguments are. But that isn't related to where they are stored.

PR #393, opened 2022-02-18
Updated return types hints and descriptions

TimothyLopas:


packages/main/src/RPA/Robocorp/Process.py
```diff
@@ -163,59 +164,63 @@ def __init__(
         self.http = HTTP()
 
     @keyword(tags=["set"])
-    def set_workspace_id(self, workspace_id: str = None):
+    def set_workspace_id(self, workspace_id: str = None) -> None:
```
cmin764, line 167:
```suggestion
    def set_workspace_id(self, workspace_id: Optional[str] = None) -> None:
```

cmin764: And for all the other similar cases :). (don't forget to `from typing import Optional` as well)

cmin764 (approved):
LGTM!

Not sure about the `None`s approach here although, like do we want to explicitly annotate them as well explicitly (with `->` and `:return(s):`) or is better to leave those alone? (might worth asking on Slack if not already agreed on this)

packages/main/src/RPA/Robocorp/Process.py
```diff
@@ -444,11 +459,12 @@ def list_process_runs_in_workspace(
     @keyword(tags=["process", "get", "runs"])
     def get_process_run_status(
         self, process_run_id: str, process_id: Optional[str] = None
-    ):
+    ) -> json:
```
mikahanninen, line 462: I think that this incorrect type when returning json objects. Better and a bit generic from json structure would be Dict[str, Any]. The json is package name.

mikahanninen:
> LGTM!
> 
> 
> 
> Not sure about the `None`s approach here although, like do we want to explicitly annotate them as well explicitly (with `->` and `:return(s):`) or is better to leave those alone? (might worth asking on Slack if not already agreed on this)

I would leave :return: None out of the doc string but keep ->

TimothyLopas:
Removed the :return: None from doc string and included Optional where applicable.

cmin764:
> Removed the :return: None from doc string and included Optional where applicable.

Cool! You can merge with latest master, push, then squash & merge.

PR #394, opened 2022-02-21
Fix `Get Os Version` to return version properly

cmin764:
Uses `platform.version()` instead of subprocess-ing `ver` command and doesn't hardcode lower than "10" versions.

### DIFF
--- a/packages/windows/poetry.lock
+++ b/packages/windows/poetry.lock
@@ -125,7 +125,7 @@ unicode_backport = ["unicodedata2"]
 
 [[package]]
 name = "click"
-version = "8.0.3"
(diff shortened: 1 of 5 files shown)


cmin764:
Test [result](https://cloud.robocorp.com/orgrobocorp/robocorp/processes/dc9842aa-ea6b-4eb6-8736-15149c8b4450/runs/c684ca54-9184-4dba-b49b-f1c336ff4263/stepRuns/21861518-051b-4dc7-9322-e5e5a728235b), look for "Running on Windows 10" log message. (tested on Windows 11 as well)

PR #395, opened 2022-02-21
Matti/move windows common to core

mmokko:
Moved the windows locator and element inspection related things from new Windows library to core.

The moved common things are needed in the inspector repository as well.

TODO: After this PR is merged the Windows library needs to inherit the context from core.
TODO: After this PR is merged the WindowsLocator keywords needs to use the WindowsElements from core.
TODO: After this PR is merged the inspector in Windows library needs to use the inspector from core.


packages/core/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "rpaframework-core"
-version = "6.5.4"
+version = "6.5.5"
```
osrjv, line 3: bump the minor for new features

packages/core/src/RPA/core/windows/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .window import Window
+from .inpect import ElementInspector
```
osrjv, line 2: `inpect`? should the file be `inspect`?

packages/core/src/RPA/core/windows/context.py
```diff
@@ -0,0 +1,116 @@
...
+from typing import Optional
+import logging
+
+from .helpers import IS_WINDOWS
+
+if IS_WINDOWS:
+    from comtypes import COMError  # noqa
+    import uiautomation as auto
+else:
+    COMError = Exception
```
osrjv, line 12: Is this really required if the uiautomation module won't be there anyways? or is this a linter thing?

packages/core/src/RPA/core/windows/context.py
```diff
@@ -0,0 +1,116 @@
...
+        self.anchor_element = None
+
+
+class WindowsContext:
+    """Shared context for all keyword libraries."""
+
+    def __init__(self, ctx):
+        if not ctx:
+            ctx = DefaultWindowsContext()
+        self.ctx = ctx
```
osrjv, line 55: very minor, but this could be: `self.ctx = ctx or DefaultWindowsContext()`

packages/core/src/RPA/core/windows/helpers.py
```diff
@@ -0,0 +1,26 @@
...
+
+def get_process_list() -> Dict:
+    """Get process list.
+
+    Returns dictionary mapping process id to process name
+    """
+    process_list = {}
+    for proc in psutil.process_iter():
+        process_list[proc.pid] = proc.name()
+    return process_list
```
osrjv, line 17: minor: dict comprehensions are a thing, e.g. `return { proc.pid: proc.name() for proc in psutil.process_iter() }`

packages/core/src/RPA/core/windows/locators.py
```diff
@@ -0,0 +1,358 @@
...
+        self.class_name = item.ClassName
+        rect = item.BoundingRectangle
+        self.left = rect.left if rect else -1
+        self.right = rect.right if rect else -1
+        self.top = rect.top if rect else -1
+        self.bottom = rect.bottom if rect else -1
+        self.width = rect.width() if rect else -1
+        self.height = rect.height() if rect else -1
+        self.xcenter = rect.xcenter() if rect else -1
+        self.ycenter = rect.ycenter() if rect else -1
```
osrjv, line 81: We could probably expose these directly instead of storing them like this. Maybe just have one `Region` object from core stored in the instance. No biggie though

osrjv (changes requested):
Ooh oops. Wasn't meant to review actual code. Rename `inpect.py` and bump a minor and I think it's fine.

packages/core/src/RPA/core/windows/context.py
cmin764: More like a linter thing, can be stripped off.
And as a nitpick, `import uiautomation as auto` can go before `from comtypes import COMError  # noqa` for a proper import sorting.

cmin764:
If it's still WIP, then discard my below comment, otherwise:

Looks legit, but I don't see where `ElementInspector` is used. If it's planned to be registered in the main `RPA.Windows` library as a `LibraryContext` derivate, then wouldn't be simpler to add just this in core and then having the windows library importing it and registering it?

I feel the concerns wouldn't be separated nicely if we move half of the windows library (destined just for it) inside a core. The inspector repo might be importing only what it needs from core then.

mmokko:
> If it's still WIP, then discard my below comment, otherwise:
> 
> Looks legit, but I don't see where `ElementInspector` is used. If it's planned to be registered in the main `RPA.Windows` library as a `LibraryContext` derivate, then wouldn't be simpler to add just this in core and then having the windows library importing it and registering it?
> 
> I feel the concerns wouldn't be separated nicely if we move half of the windows library (destined just for it) inside a core. The inspector repo might be importing only what it needs from core then (and I guess it doesn't need all these windows keywords, isn't it?).

Everything that was moved is used both in the inspector repository and in the Windows library. The Windows library uses the inspector in the main. It seems to have a recorded functionality embedded.

cmin764:
> Ooh oops. Wasn't meant to review actual code. Rename `inpect.py` and bump the minor version and I think it's fine.

Still, I haven't forgot about these... in this: https://github.com/robocorp/rpaframework/pull/402

Issue #369, opened 2022-01-27
`RPA.Desktop.Close (All) Application(s)` should really close all apps opened with this library


cmin764:
Opening with `RPA.Desktop.Open Application` apps like **Calc** or **Explorer** (which most probably spawn replacement processes) won't be closed by the keywords in the title.

Finding a way to track the resulting final process and registering that instead of the initial one in the opened processes list, can fix the problem with apps that don't get closed.

cmin764:
Researched this and can't be easily done. As for "Calculator" for example, the main process exits without letting you know what were the children and maybe the expected behaviour is to close only the main process with `Close Application` if that didn't exited automatically already.

PR #407, opened 2022-03-03
`RPA.Desktop`: Type annotations

cmin764:
No big deal with this PR, was aimed to solve this: https://github.com/robocorp/rpaframework/issues/369 (but can't be fixed and it might not even make sense to do so)


cmin764:
Nothing to test or release note about here since they're just annotations.

PR #425, opened 2022-03-17
Align core with windows (& packaging chores)

cmin764:
### ToDo
- [x] Ensure latest changes from **rpaframework-windows** are brought into **rpaframework-core** common parts as well. (+ version bump)
- [x] Make meta package infer from all the repo based sub-packages as deps. (for docs)
- [ ] Post-merge: release latest core, proceed with sister PRs of:
    A. Basing windows on latest core (removing duplicate code and use one common base from core -- suitable for: [12.10.0](https://github.com/robocorp/rpaframework/tree/release/12.10.0))
    B. Adapting inspector to latest core.

### DIFF
--- a/packages/core/src/RPA/core/windows/inspect.py
+++ b/packages/core/src/RPA/core/windows/inspect.py
@@ -14,18 +14,18 @@ class ElementInspector:
     @classmethod
     def inspect_element(
         cls,
-        verbose: bool = False,
         action: str = "Click",
         control_window: bool = True,
         recording: Optional[List["RecordElement"]] = None,
+        verbose: bool = False,
     ):
         """Inspect Windows element under mouse pointer.
 
-        :param verbose: Show exhaustive locators if `True`, otherwise just simple ones.
         :param action: Action attached to the locator.
         :param control_window: Include relevant ``Control Window  ...`` statement or
             not.
         :param recording: Where to store records.
+        :param verbose: Show exhaustive locators if `True`, otherwise just simple ones.
         """
         # TODO: Python syntax support too instead of just RF.
         output = []
@@ -66,6 +66,8 @@ class ElementInspector:
                         "top_handle": top_level_handle,
                         "x": top_level_control,
                         "locator": locator_path,
+                        "name": parent_control.Name,
+                        "control": parent_control,
                     }
                 )
 
--- a/packages/core/src/RPA/core/windows/inspect.py
+++ b/packages/core/src/RPA/core/windows/inspect.py
@@ -1,9 +1,12 @@
-from typing import List
+from typing import Dict, List, Optional, Union
 
-from .helpers import IS_WINDOWS, is_numeric
+from RPA.core.windows.helpers import IS_WINDOWS, is_numeric
 
 if IS_WINDOWS:
     import uiautomation as auto
+
+
+RecordElement = Dict[str, Union[str, auto.Control]]
 
 
 class ElementInspector:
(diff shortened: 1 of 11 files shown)


packages/core/src/RPA/core/windows/locators.py
```diff
@@ -43,9 +42,11 @@ class WindowsElement:
     """Represent Control as dataclass"""
```
cmin764, line 42: I feel this `WindowsElement` class doesn't fully work with `get_element` because the `Window` one doesn't share the same context and library construct (with `DynamicCore`) as this one. Inspector example:
1. Getting the windows elements [object](https://github.com/robocorp/inspector/blob/master/inspector/windows_driver.py#L75)
2. [Calling](https://github.com/robocorp/inspector/blob/master/inspector/windows_driver.py#L167-L168) `get_element`
3. If `"handle"` or `"executable"` are used as search params (strategy), then [listing windows](https://github.com/robocorp/rpaframework/blob/2dc3bf93b115f4c8816d48aa2b5b8b8a2c9c79c5/packages/core/src/RPA/core/windows/locators.py#L221) will fail as that method isn't found in the object instantiated at step **1.**

Solution: Following the pattern seen with the windows library [here](https://github.com/robocorp/rpaframework/blob/2dc3bf93b115f4c8816d48aa2b5b8b8a2c9c79c5/packages/windows/src/RPA/Windows/__init__.py#L28) by [registering](https://github.com/robocorp/rpaframework/blob/2dc3bf93b115f4c8816d48aa2b5b8b8a2c9c79c5/packages/windows/src/RPA/Windows/__init__.py#L488-L495) all libraries to be included.

packages/core/src/RPA/core/windows/window.py
```diff
@@ -76,14 +71,18 @@ def list_windows(
                 fullpath = win32process.GetModuleFileNameEx(handle, 0)
```
cmin764, line 71: Above we use:
```python
handle = win32api.OpenProcess(win32con.PROCESS_ALL_ACCESS, False, pid)
```
but in original windows library we [have](https://github.com/robocorp/rpaframework/blob/2dc3bf93b115f4c8816d48aa2b5b8b8a2c9c79c5/packages/windows/src/RPA/Windows/keywords/window.py#L284-L286):
```python
handle = win32api.OpenProcess(win32con.PROCESS_QUERY_LIMITED_INFORMATION, False, pid)
```
and I suspect this is because we require full info/privileges in core compared to windows which limits to minimum required access.

The question is: Do we keep separate access levels in these two (like it is right now) by defaulting to `PROCESS_ALL_ACCESS` if the windows library doesn't specify a different level? If not, shall we use `PROCESS_ALL_ACCESS` in both?

mikahanninen: I changed this in the Windows library because of principle of use least privilege needed, the `PROCESS_QUERY_LIMITED_INFORMATION` is enough to get window handle.

cmin764: So `PROCESS_QUERY_LIMITED_INFORMATION` would be just enough for both, thx!

packages/core/src/RPA/core/windows/locators.py
cmin764: Post-huddle with Matti, we like this solution, but let's try using it without importing the entire **robotframework** code:
- Maybe **robotframework-pythonlibcore** is enough, but still depends on `robot` import
- Extracting the design pattern itself from the above

cmin764: Another solution would be to just move all the methods into a single class and use them from there in the inspector, but would pose design issues when inheriting those inside the windows library.

cmin764: The winning solution remains to not depend on **robotframework** package, while having the keyword methods still split (separate concerns) and then combining all of these in a single class by inheriting the core library `WindowsEelements` class from all of them.

mmokko: That's a pretty good solution 👍.

packages/core/src/RPA/core/windows/inspect.py
```diff
@@ -1,90 +1,106 @@
...
                     {
                         "type": "locator",
-                        "top": top_level_control.Name,
-                        "top_handle": top_level_control.NativeWindowHandle,
+                        "top": top_level_name,
+                        "top_handle": top_level_handle,
                         "x": top_level_control,
                         "locator": locator_path,
-                        "name": parent_control.Name,
-                        "control": parent_control,
```
mmokko, line 54: I think the control is needed in the inspector side.

mmokko (changes requested):
Looks good to me 👍. Check the return type for the recording, it should match what the inspector is expecting in the windows_driver.py on_click function.

packages/core/src/RPA/core/windows/inspect.py
cmin764: You're right, good catch!

cmin764: Yes, good catch!

Issue #404, opened 2022-03-01
Base `RPA.Windows` on core


cmin764:
For convenience and better experience, we decided to include the new beta windows library into the main one as a referenced dependency.

### ToDo
- [x] 1. Sort out icons [PR](https://github.com/robocorp/rpaframework/pull/401) and up-to-date base from `RPA.Windows`. (and release latest _core_)
- [x] 2. Make `RPA.Windows` import common stuff from `RPA.core` while removing duplicate code.
- [ ] 3. Release latest _windows_ and link this dependency in _main_. (so _rpaframework_ brings latest _rpaframework-windows_ automatically -- fix docs meta-package deps)
- [x] 4. Add deprecation message in `RPA.Desktop.Windows` in favor of using `RPA.Windows` instead. (then release _main_)

cmin764:
@mmokko / @mikahanninen Let me know if that's the right plan.

PR #431, opened 2022-03-21
Release/12.10.0

mikahanninen:
Updated libraries (check the box after testing)
- [x] RPA.Cloud.AWS (fix. Download Files saving objects with paths)
- [x] RPA.Robocorp.Process (changed. type hints, doc strings)
- [x] RPA.HTTP (fix. override RequestsLibrary logging request and response to DEBUG level)
- [x] `RPA.Windows` (**core** basing; optional: ~extra attributes when listing processes~ (this would get into **core** now because there we have the listing windows functionality now), clicking with coords) (Issues: #404 ) -> moved from here into #435

Tasks
- [x] test
- [x] update releasenotes
- [ ] release


pyproject.toml
```diff
@@ -9,12 +9,7 @@ authors = [
 [tool.poetry.dependencies]
 python = "^3.6.2"
 rpaframework = { path = "packages/main", extras = ["cv", "playwright", "aws"] }
-rpaframework-core = { path = "packages/core" }
-rpaframework-dialogs = { path = "packages/dialogs" }
 rpaframework-google = { path = "packages/google" }
-rpaframework-pdf = { path = "packages/pdf" }
-rpaframework-recognition = { path = "packages/recognition" }
-rpaframework-windows = { path = "packages/windows" }
```
cmin764, line 13: I removed these since they are already included by **main** and we would have package version resolution issues since the latest code in _master_ isn't necessarily compatible with all the pinned versions.

The docs should still compile with the latest **main** in _master_ (and its pins) + any other non-**main** referenced sub-package we have.

cmin764:
This is how the warning looks like now:

Mac
nix-warn


Windows
win-warn

PR #433, opened 2022-03-22
Bump every package to core 7.0.0

cmin764:
- **core** bumped to `7.0.0` since would be a breaking change for the current inspector
- since **windows** is based on latest **core** and **main** is based on **windows** and **recognition**, we require to get the same `7.0.0` **core** for **recognition** as well
- and **meta** implies **main** and the rest of the sub-packages, which all should depend on the latest `7.0.0` **core**

### DIFF
--- a/.github/workflows/dialogs.yaml
+++ b/.github/workflows/dialogs.yaml
@@ -26,7 +26,7 @@ jobs:
       fail-fast: false
       matrix:
         name: [
-          "windows-py36",
+          # "windows-py36", no support anymore for the package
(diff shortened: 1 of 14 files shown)


mikahanninen (approved):

cmin764:
I manually ran `build` _invoke_ command under all packages:
- mac: dialogs, pdf, google
- windows: recognition, windows

And linting, testing and building just passed. (not pubished yet ofc, just to Devpi dev so far)

PR #453, opened 2022-04-05
`RPA.Database`: Normalize rows to list of tuples

cmin764:
Ensures rows is a list of tuples before returning or formatting them into a table, otherwise we might end up with entire row data "squeezed" into the first column.

### DIFF
--- a/docs/source/releasenotes.rst
+++ b/docs/source/releasenotes.rst
@@ -8,6 +8,12 @@ Release notes
 `Released <https://pypi.org/project/rpaframework/#history>`_
 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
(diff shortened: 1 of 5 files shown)


cmin764:
Tested with robot:
```robotframework
Query my database test
    connect to database     module_name=pyodbc  database=db14215456
    ...                     username=lucielavickovaDEV  password=sk**********55
    ...                     host=localhost\\SQLEXPRESS
    @{query_result}    query    SELECT firstname, lastname, email, active FROM Users
    log  ${query_result}
    disconnect from database
```
through Assistant and got:
Screenshot 2022-04-05 at 16 44 40
Screenshot_2022-04-05_at_16_44_58

so all works correctly now. (including auth to SQL Server 2019)

PR #455, opened 2022-04-06
Next button for wizard in RPA.Dialogs

### LINKED ISSUE #452: Wizard-like forms in RPA.Dialogs
Add support for:

- [ ]  New keyword `Add dialog next page button`

This keyword will support a `label` argument and will add a button that separates dialog content between multiple views, like a wizard.

bogdancondurache:
Closes #452 

Adds support for the `Add dialog next page button` keyword.

### DIFF
--- a/docs/source/releasenotes.rst
+++ b/docs/source/releasenotes.rst
@@ -5,7 +5,7 @@ Release notes
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
 - Library **RPA.Database**: Fix configuration value retrieval. (:pr:`456`)
-- Library **RPA.Dialogs**: Add next button to support wizard style dialogs. (:pr:`455`)
+- Library **RPA.Dialogs**: Add next button to support wizard style dialogs. (:issue:`452`)
 
 
 `Released <https://pypi.org/project/rpaframework/#history>`_
--- a/docs/source/releasenotes.rst
+++ b/docs/source/releasenotes.rst
@@ -5,6 +5,7 @@ Release notes
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
 - Library **RPA.Database**: Fix configuration value retrieval. (:pr:`456`)
+- Library **RPA.Dialogs**: Add next button to support wizard style dialogs. (:pr:`455`)
 
 
 `Released <https://pypi.org/project/rpaframework/#history>`_
--- a/packages/core/pyproject.toml
+++ b/packages/core/pyproject.toml
@@ -43,7 +43,7 @@ pillow = "^8.4.0"
 [tool.poetry.dev-dependencies]
 black = {version = "^21.5b0", allow-prereleases = true}
 flake8 = "^3.7.9"
-pylint = "^2.4.4"
+pylint = "^2.4.4, <2.13"
 pytest = "^6.1.2"
 pytest-cov = "^2.10.1"
 mock = "^4.0.2"
--- a/packages/dialogs/src/RPA/Dialogs/library.py
+++ b/packages/dialogs/src/RPA/Dialogs/library.py
@@ -830,8 +830,9 @@ class Dialogs:
 
         :param label: The text displayed on the button
 
-        The next page button will delimit the next elements of the dialog in a separate view,
-        so that they will only be displayed after pressing the button, creating a wizard experience in the dialog.
+        The next page button will delimit the next elements of
+        the dialog in a separate view, so that they will only be displayed after
+        pressing the button, creating a wizard experience in the dialog.
 
         Example:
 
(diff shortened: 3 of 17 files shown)


packages/dialogs/src/RPA/Dialogs/library.py
```diff
@@ -821,6 +821,44 @@ def add_checkbox(
 
         self.add_element(element)
 
+    @keyword("Add dialog next page button")
+    def add_dialog_next_page_button(
+        self,
+        label: str,
+    ) -> None:
+        """Next button todo
```
cmin764, line 829: I guess this `"todo"` is still WIP.

packages/dialogs/src/RPA/Dialogs/library.py
```diff
@@ -821,6 +821,44 @@ def add_checkbox(
...
+
+            Add heading       Send feedback
+            Add text input    email             label=E-mail address
+            Add dialog next page button         label=next
+            Add text input    name              label=Name
+            Add dialog next page button         label=next
+            Add checkbox      name=contact
+            ...     label=Do you want us to contact you?
+            ...     default=True
+            Add dialog next page button         label=next
```
cmin764, line 845: Cool!

packages/dialogs/tests/test_dialogs.robot
```diff
@@ -72,6 +72,22 @@ All elements
...
+    Add heading    2/3
+    Add text input      text-field-2
+    Add dialog next page button         label=next2
+
+    Add heading    3/3
+    Add text input      text-field-3
+    Add submit buttons  submit
+
+    ${result}=  Run dialog
+    Log many    &{result}
```
cmin764, line 89: Nice idiom!

cmin764 (approved):
Good job, lgtm!

One thing that's left to be added: some release notes.

You can do it in the _Upcoming_ section and release later on, alongside other changes in another PR where you move the release note from the _Upcoming_ to the _Released_ incremented version. (including `rpaframework-dialogs` package updated pin alongside `rpaframework` (main package) version increment) -- see [here](https://github.com/robocorp/rpaframework/pull/456/files) an example

packages/core/pyproject.toml
```diff
@@ -43,7 +43,7 @@ pillow = "^8.4.0"
 [tool.poetry.dev-dependencies]
 black = {version = "^21.5b0", allow-prereleases = true}
 flake8 = "^3.7.9"
-pylint = "^2.4.4"
+pylint = "^2.4.4, <2.13"
```
cmin764, line 46: If you don't want to have "fun" with releasing all our packages, maybe pinning this for now into dialogs and main packages should be enough. (but I guess you know best and it's even better to fix this problem everywhere else it happens)

Good luck!

bogdancondurache: well, I think this is just a version pin and shouldn't be released; it's just for the developer experience so that whoever works on those projects in the future won't have the pipeline failing. Is there a reason why a release would be necessary for that?

packages/dialogs/src/RPA/Dialogs/library.py
bogdancondurache: yes, forgot to update it, thanks

packages/core/pyproject.toml
cmin764: May bad, since it's dev only then there's no reason to do that as this gets used with `poetry install`s only. (but never published afaik)

cmin764: May bad, since it's dev only then there's no reason to do that as this gets used with `poetry install`s only. (but never published afaik)

docs/source/releasenotes.rst
```diff
@@ -5,6 +5,7 @@ Release notes
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
 - Library **RPA.Database**: Fix configuration value retrieval. (:pr:`456`)
+- Library **RPA.Dialogs**: Add next button to support wizard style dialogs. (:pr:`455`)
```
cmin764, line 8: Here you can put your issue like so:
```suggestion
- Library **RPA.Dialogs**: Add next button to support wizard style dialogs. (:issue:`452`)
```

Forgot to mention that issues take precedence in front of PRs. (if an issue exists ofc, for simpler hotfixes sometimes we do PRs directly, but never pushing directly to master :D)

cmin764 (approved):
All good!

packages/dialogs/src/RPA/Dialogs/library.py
```diff
@@ -821,6 +821,48 @@ def add_checkbox(
 
         self.add_element(element)
 
+    @keyword("Add dialog next page button")
```
mikahanninen, line 824: It is unnecessary declare keyword name in the decorator as name is the same as method name.

bogdancondurache: ah, didn't know that, thanks 😄 I went based on examples and all the keywords in RPA.Dialogs also had the keyword name in the decorator. Removed

bogdancondurache, line 824:
```suggestion
    @keyword
```

PR #461, opened 2022-04-07
Create new RFW Library `RPA.Hubspot`

blakewaud:
This is the replacement for #438. That PR's commit history was muddled by my own mischief.

### PR Contents Description

The RPA.Hubspot library currently focuses on getting data from the CRM module of Hubspot. The underlying python library hubspot-api-client is an automatically generated SDK from [OpenAPiTools](https://github.com/OpenAPITools/openapi-generator) and has many more commands available that I have not yet been able to wrap for RFW.

The biggest feature of this API is a "natural language" search mode for the search API whereby a RFW developer can write a query like so:

``` robotframework
    ${contacts}=    Search for objects    CONTACTS
    ...    firstname    EQ    John    AND    lastname    EQ    Smith
```

Currently, the library returns found objects as the native object or list of those native objects. This allows a user to access many properties of the Hubspot objects via dot notation. Some of the underlying objects are dictionaries (most notably the `properties` attribute of `SimplePublicObject` types).


packages/main/src/RPA/Hubspot.py
```diff
@@ -0,0 +1,1706 @@
...
+
+        return collected_responses
+
+    @keyword
+    @retry(
+        retry=retry_if_exception(_is_rate_limit_error),
+        stop=stop_after_attempt(10),
+        wait=wait_exponential(multiplier=2, min=0.1),
+        before_sleep=_before_sleep_log(),
+    )
```
cmin764, line 1234: Not sure if this should be retried as well since you're already retrying the underlying calls into `_get_object_by_batch` and `_get_object`.

packages/main/pyproject.toml
```diff
@@ -2,9 +2,7 @@
 name = "rpaframework"
 version = "13.1.0"
```
cmin764, line 3: Maybe we should increase library version now or in another subsequent PR once we decide to release this.

cmin764 (approved):
Didn't tested it at all, just took a quick glance and looks in very good shape now!

I think some _Upcoming_ release notes would be good here... or even right under _Released_ if you plan to release immediately afterwards.

packages/main/src/RPA/Hubspot.py
blakewaud: No, that's redundant, thanks!

PR #470, opened 2022-04-14
Add Robot Framework v5 support

mikahanninen:
Gathering test results on [Robot Framework version 5](https://coda.io/d/Robot-Framework-development_dGUWyFjAURP/Robot-Framework-version-5_sutUm#Bot-results_tu8RV/r1) Coda.

### ToDo
- [x] `poetry update` for all sub-packages
- [x] Run all unit-test (pytest)
- [x] Run all RF tests (robot)


cmin764 (approved):
LGTM!

Also checked on the latest RF5 `deco` module as we have it [vendorized](https://github.com/robocorp/rpaframework/blob/master/packages/core/src/RPA/core/vendor/deco.py) and there was an extra `library` argument added called `converters`, decorator which we don't use from the vendorized file (as we use the `keyword` one only on Windows related common methods):
Screenshot 2022-04-19 at 01 26 03

PR #484, opened 2022-04-29
Implement `RPA.Hubspot` update keywords

blakewaud:
Creates keywords necessary to update data in Hubspot, include creating and updating objects. Batched inputs to these keywords require special handling because of how the underlying library works so I borrowed the pattern of having a single working object that the library tracks and the user can manipulate until they are ready to execute the batch.

### DIFF
--- a/packages/main/src/RPA/Hubspot.py
+++ b/packages/main/src/RPA/Hubspot.py
@@ -289,11 +289,14 @@ class BatchInputFactory:
         self,
         mode: BatchMode = BatchMode.INVALID,
         object_type: str = None,
-        inputs: List = [],
+        inputs: List = None,
     ) -> None:
         self._mode = mode
         self._object_type = object_type
-        self._inputs = inputs
+        if inputs:
+            self._inputs = inputs
+        else:
+            self._inputs = []
 
     def __len__(self) -> int:
         return len(self.inputs)
@@ -425,7 +428,7 @@ class Hubspot:
     the nearest integar to include in API calls (i.e., the resulting
     integer sent to the API must have 13 digits as of March 18, 2022).
 
-    Example usage:
+    **Robot framework example usage:**
 
     .. code-block:: robotframework
 
@@ -441,6 +444,8 @@ class Hubspot:
             ${deals}=    Search for objects    DEALS
             ...    hs_lastmodifieddate    GTE    ${yesterday_hs_ts}
 
+    **Python example usage**
+
     .. code-block:: python
 
         from robot.libraries.DateTime import get_current_date, subtract_time_from_date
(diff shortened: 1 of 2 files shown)


blakewaud:
This does produce some divergence in how batched gets work vs the new batched inputs. An improvement down the road would be to expand `BatchModes` to include `GET` and `DELETE`.

packages/main/src/RPA/Hubspot.py
```diff
@@ -257,9 +270,160 @@ def create_search_object(self, words: List[str] = None):
...
+    def extend_inputs(
+        self, properties: List[Dict[str, str]], ids: List[str] = None
+    ) -> None:
+        """Extends this batch's ``inputs`` based on the provided list of
+        ``properties`` and ``ids``. If ``ids`` is provided, it will be
+        zipped with the provided ``properties``. ``UPDATE`` mode is assumed
+        in such a case.
+        """
+        if ids:
+            if len(ids) > 0:
```
cmin764, line 394: I feel this `if ids` already ensures the condition of having one or more elements, so why double check?

packages/main/src/RPA/Hubspot.py
```diff
@@ -309,6 +475,83 @@ class Hubspot:
...
+    but when wishing to create or update many objects, the library provides
+    a batching system.
+
+    In order to start a batch, you must first call the \`Create new batch\`
+    keyword. This initializes a new batch to accept inputs. If a batch
+    already exists when you call this keyword, it will be lost and a new
+    blank one will be started.
+
+    Once the batch has been initialized, you can add inputs one at a time with
+    \`Add input to batch\` or many at a time with \`Extend batch with inputs\`.
```
cmin764, line 494:
```suggestion
    ``Add input to batch`` or many at a time with ``Extend batch with inputs``.
```
in REST docstrings (or in [release notes](https://rpaframework.org/releasenotes.html)) we usually use double '``' to highlight a piece of keyword/code. (as single '`' just emphasises it like single '*')

This way you won't need anymore to use repr `r"""...` docstrings anymore.

cmin764 (changes requested):
Left a few nits and a bug, overall looks good!

Would be awesome if you can add some _pytest_ unit-tests to atomically test a few things that don't rely on the API key, but I'm not insisting on it. (fake data through mocks)

General nit: setters/getters properties are cool, but make sure you use them if and only if you do something extra/more rather than just setting a value to a "private" attribute. (overall is ok, but not sure if really required for `object_type` and `batch_input`)

packages/main/src/RPA/Hubspot.py
```diff
@@ -257,9 +270,160 @@ def create_search_object(self, words: List[str] = None):
...
+        mode: BatchMode = BatchMode.INVALID,
+        object_type: str = None,
+        inputs: List = None,
+    ) -> None:
+        self._mode = mode
+        self._object_type = object_type
+        if inputs:
+            self._inputs = inputs
+        else:
+            self._inputs = []
```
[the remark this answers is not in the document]
blakewaud: Didn't know this pattern worked in Python, thank you!

packages/main/src/RPA/Hubspot.py
```diff
@@ -257,9 +270,160 @@ def create_search_object(self, words: List[str] = None):
...
+        return len(self.inputs)
+
+    @property
+    def mode(self) -> BatchMode:
+        """The batch mode, either ``UPDATE`` or ``CREATE``."""
+        return self._mode
+
+    @mode.setter
+    def mode(self, mode: BatchMode):
+        if mode not in [BatchMode.UPDATE, BatchMode.CREATE]:
```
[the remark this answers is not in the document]
blakewaud: Ah, yeah, that's what I wanted to do so that's much better.

packages/main/src/RPA/Hubspot.py
blakewaud: Because it was a hold over when I refactored it, you are right.

packages/main/src/RPA/Hubspot.py
blakewaud: By escaping the backtick, it actually invoked a special component of the Robot Framework docgen tool and it creates an inter keyword link, that's why I do it.

blakewaud:
> Would be awesome if you can add some _pytest_ unit-tests to atomically test a few things that don't rely on the API key, but I'm not insisting on it. (fake data through mocks)

Yes, that's something I had intended to do once I better understood the actual API but didn't get around to it on my first pass because I was focusing on testing against sandbox. Do you know of a good example in our code base doing mocks currently, @cmin764 ? I will endeavor to add this for next time after I get a bot this is waiting on to work.

> General nit: setters/getters properties are cool, but make sure you use them if and only if you do something extra/more rather than just setting a value to a "private" attribute. (overall is ok, but not sure if really required for `object_type` and `batch_input`)

Fair point, in that case, do you just define it as `self.object_type` in `__init__` and it works the same? Does it show up as a known attribute in intellisense when you import it somewhere else?

I think I'll keep `batch_input` how it is since I do have some code on the getter and deleter.

> 1. Q: Are the robot tests also automated in CI (like you being able to provide the secret API key) or are you running them manually before merging?

It is automated in CI via a secret that gets injected as the variable file.

cmin764:
- For mocks example `grep -rni "mock" packages/main/tests/python` will help you find some good examples, like _packages/main/tests/python/test_robocorp_workitems.py_ for example, look for the usage of the `mock` library and how patching works. (also _pytest_ docs are a great source)
- On properties over-engineering: Exactly! Use "public" attributes directly like you described. (there's a convention in Python where what's used outside too, it's "public" and for internal access only is "private" -- beginning with an underscore `_`; you can use `__` for [name-mangling](https://www.geeksforgeeks.org/name-mangling-in-python/) if you really want to alter behaviour)
- On the CI testing: Nice!

PR #488, opened 2022-05-04
Add file stem keyword to file system

vx-i:
There is a keyword for file name and suffix, but file name stem keyword is missing.


cmin764:
> There is a keyword for file name and suffix, but file name stem keyword is missing.

Indeed, that looks useful when you just want the name without coming with your own algorithm of extension stripping. Transferred changes to our local branch based PR: https://github.com/robocorp/rpaframework/pull/597

PR #490, opened 2022-05-04
Create extended excel file examples

TimothyLopas:
Extended the RPA.Excel.File Robot Framework examples and added Python examples.


packages/main/src/RPA/Excel/Files.py
```diff
@@ -468,15 +738,17 @@ def find_empty_row(self, name=None):
...
 
         .. code-block:: robotframework
 
-            Append rows to worksheet     ${rows}
             ${next}=    Find empty row
-            Insert image to worksheet    ${next}    A    screenshot.png
+
+        .. code-block:: python
+
+            next = lib.find_mpty_row()
```
cmin764, line 751:
```suggestion
            next = lib.find_empty_row()
```

cmin764 (changes requested):
Found a dummy typo, but looks good!

Bonus: Would be hard to build a robot for yourself where you can gather and test all the documented keyword examples under the docstrings just to make sure they are working as expected? (we can skip it for the Py side)

cmin764 (approved):
I guess it looks alright now, thx!

PR #491, opened 2022-05-05
Excel: Files: Fix issues with removing worksheets

osrjv:
- Fix IndexError when removing (certain) worksheets from XLS workbooks
- Reset active worksheet if it is deleted
- Ensure the last worksheet can not be deleted

Original issue reported internally by @TimothyLopas

### DIFF
--- a/packages/main/src/RPA/Excel/Files.py
+++ b/packages/main/src/RPA/Excel/Files.py
@@ -787,6 +787,14 @@ class XlsxWorkbook:
 
     def remove_worksheet(self, name=None):
         name = self._get_sheetname(name)
+        others = [sheet for sheet in self.sheetnames if sheet != name]
+
+        if not others:
+            raise ValueError("Workbook must have at least one other worksheet")
+
+        if name == self.active:
+            self.active = others[0]
+
         sheet = self._book[name]
         self._book.remove(sheet)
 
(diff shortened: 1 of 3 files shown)


packages/main/src/RPA/Excel/Files.py
```diff
@@ -787,6 +787,14 @@ def append_worksheet(self, name=None, content=None, header=False, start=None):
 
     def remove_worksheet(self, name=None):
         name = self._get_sheetname(name)
+        others = [sheet for sheet in self.sheetnames if sheet != name]
+
+        if not others:
+            raise ValueError("Workbook must have at least one worksheet")
```
mikahanninen, line 793: Just curious - why there needs to be this kind of protection? I mean why user can't just delete all sheets and create then new sheet.

osrjv: At least .xls workbooks (with the upstream library) can not have zero worksheets, it breaks in spectacular ways. With .xlsx that can be done, as long as the user adds one before saving. Not sure if I want the keywords to behave differently based on the file extension.

mikahanninen: Yeah that is fine then to keep the protection. Thanks 👍

cmin764, line 793: Nit
```suggestion
            raise ValueError("Workbook must have at least one other worksheet")
```
if that makes more sense, so the user understands that his one and only worksheet isn't enough as it leaves the Excel empty.

cmin764 (approved):
LGTM!

Also, it is possible to link in this PR the reported issue? (if that's created already)

packages/main/src/RPA/Excel/Files.py
osrjv: Fixed in both places

PR #492, opened 2022-05-06
RPA.Tables Examples Updated

TimothyLopas:


docs/source/releasenotes.rst
```diff
@@ -19,6 +19,7 @@ Release notes
 
 - Library **RPA.Excel.Files** (:pr:`490`): Keyword examples updated to be more complete
   and Python examples have been added to all keywords.
+- Library **RPA.Tables** (:pr:`492`): Keyword examples updated to be more complete
```
cmin764, line 22: I guess this has to be updated with the latest _master_ since we released **14.1.1** meanwhile. (and the diff isn't under the **Upcoming** section anymore I guess)

packages/main/src/RPA/Tables.py
```diff
@@ -1241,7 +1269,7 @@ def add_table_row(self, table: Table, values: Any = None):
         It can also be a single value that is set for all columns,
         which is ``None`` by default.
 
-        Example:s
+        Examples:
```
cmin764, line 1272: You have a sharp eye! :D

packages/main/src/RPA/Tables.py
```diff
@@ -1466,14 +1492,14 @@ def table_head(
...
 
-        Examples:
+        Example:
 
         .. code-block:: robotframework
 
             # Get the first 10 employees
-            ${employees}=    Read worksheet as table    employees.xlsx
-            ${first}=        Table head    ${employees}    10
+            ${first}=    Table head    ${employees}    10
```
cmin764, line 1502: Is it ok to hide the source of `${employees}`? Seen the same idiom at this [line](https://github.com/robocorp/rpaframework/pull/492/files#diff-25a3bbc35959efc76efba5321b5512ceb20cd1cc4f8896beeaf4b333480a8eb1L1455) as well.

Not a big deal, just curious if it makes more sense in the docs to focus on the keyword effect itself. (rather than including its setup as well)

cmin764 (approved):
Looks good, nothing strange spotted!

Just curious why do we remove the pre-setup part from the examples. (like when we're stripping the part of getting the initial `${orders}` or `${employees}` we use later on to manipulate the data with)

packages/main/src/RPA/Tables.py
TimothyLopas: I'm hiding this detail as our experienced users understand where this comes from and my experience with new users is that they blindly copy and paste the examples and think that this is the only way for this to work. To avoid that I wanted to keep our examples more generic and let them determine how to bring in the table.

docs/source/releasenotes.rst
TimothyLopas: Updated with the latest master

TimothyLopas: Updated to latest version of master

cmin764 (approved):
LGTM!

Fix the conflicts, then merge!

PR #493, opened 2022-05-06
RPA.Excel.Files typehint updates

TimothyLopas:
@cmin764 There was a merge conflict with PR 491 that I resolved. Please confirm that I resolved it correctly and did not reverse any of the PR 491 updates.


packages/main/src/RPA/Excel/Files.py
```diff
@@ -198,7 +201,9 @@ def _load_workbook(self, path, data_only):
             "verify that the path and extension are correct"
         )
 
-    def create_workbook(self, path=None, fmt="xlsx"):
+    def create_workbook(
+        self, path: Optional[str] = None, fmt: Optional[str] = "xlsx"
```
cmin764, line 205:
```suggestion
        self, path: Optional[str] = None, fmt: str = "xlsx"
```
wouldn't break unexpectedly if you pass `fmt=None` (will raise `ValueError(f"Unknown format: None")`) but also I see no point on marking it as `Optional` since it's not like you're getting any default good behaviour if you send a `None` to this param. (remember, there's a [diff](https://docs.python.org/3/library/typing.html#typing.Optional) between optional parameters (keyword arguments) and `Optional` values)

> Note that this is not the same concept as an optional argument, which is one that has a default. An optional argument with a default does not require the Optional qualifier on its type annotation just because it is optional. For example:

packages/main/src/RPA/Excel/Files.py
```diff
@@ -406,10 +421,11 @@ def list_worksheets(self):
         assert self.workbook, "No active workbook"
         return self.workbook.sheetnames
 
-    def worksheet_exists(self, name):
+    def worksheet_exists(self, name: str) -> bool:
         """Return True if worksheet with given name is in workbook.
 
         :param name: Name of worksheet you are looking for
+        :return: `True` if the worksheet exists; else `False`
```
cmin764, line 428: Nit
```suggestion
        :return: `True` if the worksheet exists, `False` otherwise
```

packages/main/src/RPA/Excel/Files.py
```diff
@@ -582,7 +613,13 @@ def read_worksheet(self, name=None, header=False, start=None):
         assert self.workbook, "No active workbook"
         return self.workbook.read_worksheet(name, header, start)
 
-    def read_worksheet_as_table(self, name=None, header=False, trim=True, start=None):
+    def read_worksheet_as_table(
+        self,
+        name: Optional[str] = None,
+        header: Optional[bool] = False,
+        trim: Optional[bool] = True,
```
cmin764, line 620:
```suggestion
        header: bool = False,
        trim: bool = True,
```
It's not commonly expected to pass `None`s to booleans like these. Usually you rely on its default value (by not specifying the param during the call) or send `True`/`False` explicitly, therefore no `Optional` annotation.

packages/main/src/RPA/Excel/Files.py
```diff
@@ -623,14 +661,21 @@ def read_worksheet_as_table(self, name=None, header=False, trim=True, start=None
         sheet = self.read_worksheet(name, header, start)
         return tables.create_table(sheet, trim)
 
-    def append_rows_to_worksheet(self, content, name=None, header=False, start=None):
+    def append_rows_to_worksheet(
+        self,
+        content: Any,
+        name: Optional[str] = None,
+        header: Optional[bool] = False,
```
cmin764, line 668:
```suggestion
        header: bool = False,
```

packages/main/src/RPA/Excel/Files.py
```diff
@@ -857,7 +919,14 @@ def set_cell_format(self, row, column, fmt, name=None):
         assert self.workbook, "No active workbook"
         self.workbook.set_cell_format(row, column, fmt, name)
 
-    def insert_image_to_worksheet(self, row, column, path, scale=1.0, name=None):
+    def insert_image_to_worksheet(
+        self,
+        row: int,
+        column: Union[int, str],
+        path: str,
+        scale: Optional[float] = 1.0,
```
cmin764, line 927:
```suggestion
        scale: float = 1.0,
```
scale should be a `float` (or number) at all times.

cmin764 (changes requested):
Definitely it brings more value to have all of these annotations!

Just watch out for those `Optional` ones; as a rule of thumb we mark as `Optional` those params whose values can be `None`s as well, not the optional params (keyword arguments with default values) -> which in most cases expect values of the same type as the default value they hold.

[Read more](https://docs.python.org/3/library/typing.html#typing.Optional)

cmin764 (approved):
LGTM! Wouldn't hurt to have a simple release note just to tie this PR to it under the notes. (and I guess this is covered already by #492, so we're good)

Issue #220, opened 2021-09-10
Documentation: Find Table Rows examples missing the table argument


emanlove:
The examples under the [Find table Rows keyword doc](https://rpaframework.org/libdoc/RPA_Tables.html#Find%20Table%20Rows) are as shown,

```robotframework
# Find all rows where price is over 200
@{rows}=    Find table rows    Price  >  ${200}

# Find all rows where the status does not contain "removed"
@{rows}=    Find table rows    Status    not contains    removed
```

are missing the **table** argument, which for completeness one might add the line `${orders}=    Read worksheet as table    orders.xlsx` so as to have a table as compared to the generic example `${table}`

PR #495, opened 2022-05-09
`RPA.Tables`: Add keywords for mapping column values and removing rows with custom filters

osrjv:
Long-standing branch that was never merged, but has some valid use-cases. Allows running an arbitrary keyword for each cell of a column to tranform values between types or create mappings.


osrjv:
Added another keyword, which allows filtering with custom keywords

osrjv:
Added another keyword, which allows filtering with custom keywords

cmin764:
Closes #220

Issue #501, opened 2022-05-11
`RPA.Browser` auto-add URL scheme if missing


cmin764:
Automatically prepend URL scheme (`https://` protocol) if missing in the URLs sent as arguments in the following keywords:
- [Open Browser](https://robocorp.com/docs/libraries/rpa-framework/rpa-browser-playwright/keywords#open-browser)
- [Open Available Browser](https://robocorp.com/docs/libraries/rpa-framework/rpa-browser/keywords#open-available-browser)
- [Open Chrome Browser](https://robocorp.com/docs/libraries/rpa-framework/rpa-browser/keywords#open-chrome-browser)
- [Open User Browser](https://robocorp.com/docs/libraries/rpa-framework/rpa-browser-selenium/keywords#open-user-browser)

For any of the Selenium/Playwright flavours if the absence of such scheme produces an error (during the implicit redirect).

PR #502, opened 2022-05-11
Selenium: Ensure URL has a protocol when navigating

osrjv:
Added an override to the `Go To` method/keyword, which autoformats URLs by default. At this point it means appending an `https://` protocol if none is given. This fixes the direct usage, but also indirect through `Open Available Browser`.

It does _not_ fix the usage through plain `Open Browser`, which manipulates the driver object directly. However, due to how DynamicCore and inheritance works, there is no clean way to wrap parent keywords. It would have to be a look-up through the `self.keywords` attribute by string name.

NB: In addition to `http://`, `https://`, `ftp://` type protocols, it should also not prefix when giving browser internal destinations such as `about:config`.


packages/main/src/RPA/Browser/Selenium.py
```diff
@@ -1949,3 +1950,19 @@ def _send_command_and_get_result(self, cmd, params):
...
+    @keyword
+    def go_to(self, url: str, autoformat=True):
+        """Navigates the current browser window to the provided ``url``.
+
+        If the ``autoformat`` argument is enabled (default), the
+        library adds an ``https://`` prefix if no protocol is given.
+        """
+        if autoformat:
+            url = str(url).strip()
+            if not re.match(r"^\w+:", url):
```
cmin764, line 1963: While this indeed works, wouldn't be safer to use `urlsplit/urlparse` from `urllib.parse` like so?
```python
>>> urllib.parse.urlsplit("https://www.google.com/path/to?q=1")
SplitResult(scheme='https', netloc='www.google.com', path='/path/to', query='q=1', fragment='')
>>> urllib.parse.urlsplit("about:config").scheme
'about'
```

packages/main/src/RPA/Browser/Selenium.py
```diff
@@ -1949,3 +1950,19 @@ def _send_command_and_get_result(self, cmd, params):
...
+                self.logger.debug("Adding missing protocol to URL (https://)")
+                url = "https://" + url
```
cmin764, line 1965: And then you can use `urllib.parse.urlunsplit(...)` over the listed parts above where you just replace the `scheme` with `https`. ([example](https://github.com/robocorp/example-oauth-email/blob/master/OAuth2.py#L16))

packages/main/src/RPA/Browser/Selenium.py
```diff
@@ -1949,3 +1950,19 @@ def _send_command_and_get_result(self, cmd, params):
...
+
+        self.logger.info(f"Opening url '{url}'")
+        self.driver.get(url)
```
cmin764, line 1968: Nit:
```suggestion
        super().go_to(url)
```

cmin764 (approved):
LGTM, no issues observed, but still left a few potential improvements.

Also, why is this _partially_ fixing the [issue](https://github.com/robocorp/rpaframework/issues/501)? Is it because this PR is addressing only the Selenium fix and we remain with the Playwright one? If "yes", would Playwright work just fine with scheme-less URLs? (if "yes", then we can fully close the issue afterwards)

packages/main/src/RPA/Browser/Selenium.py
osrjv: I referred to this in the PR description. This does not work, since there is no direct `go_to` method in the super class. The upstream library uses [DynamicCore](https://github.com/robotframework/PythonLibCore) which does some trickery around how the methods are registered/called.

osrjv:
> LGTM, no issues observed, but still left a few potential improvements.
> 
> Also, why is this _partially_ fixing the [issue](https://github.com/robocorp/rpaframework/issues/501)? Is it because this PR is addressing only the Selenium fix and we remain with the Playwright one? If "yes", would Playwright work just fine with scheme-less URLs? (if "yes", then we can fully close the issue afterwards)

Partially because the keyword `Open Browser` does not use this method internally, but instead manipulates the Selenium webdriver object directly. We would have to inherit the keyword and then override the `url` argument formatting, but it's a bit annoying because of the reason I mentioned in the other comment (and in OP). Maybe I should still try to make a fix for it.

osrjv:
Changed it a bit to take into account review comments.
- Using urllib.parse for split/unsplit
- Added a keyword for setting the default scheme, to allow changing the value for nested calls

packages/main/src/RPA/Browser/Selenium.py
```diff
@@ -54,10 +56,73 @@ def suppress_logging():
         logger.warning, logger.error = logger_warning, logger_error
 
 
+def ensure_scheme(url, default):
```
cmin764, line 59: Type annotations wouldn't hurt here as well.

packages/main/src/RPA/Browser/Selenium.py
```diff
@@ -54,10 +56,73 @@ def suppress_logging():
...
+        value, such as `http` or `https`.
+        """
+        self._default_scheme = scheme
+
+    @keyword
+    def go_to(self, url: str) -> None:
+        url = ensure_scheme(url, self._default_scheme)
+        super().go_to(url)
+
+    go_to.__doc__ = BrowserManagementKeywords.go_to.__doc__
```
cmin764, line 95: Does its job, but if you feel this is even more elegant, then go for it (not tested):
```python
@keyword
@functools.wraps(BrowserManagementKeywords.go_to)
def go_to(...)
```

packages/main/src/RPA/Browser/Selenium.py
```diff
@@ -405,6 +470,11 @@ def __init__(self, *args, **kwargs) -> None:
         kwargs["plugins"] = ",".join(plugins)
 
         SeleniumLibrary.__init__(self, *args, **kwargs)
+
+        # Add inherit/overriden library keywords
+        overrides = [BrowserManagementKeywordsOverride(self)]
+        self.add_library_components(overrides)
```
cmin764, line 476: Perfect! Glad this just works.

cmin764 (approved):
Definitely this is better, some last nits left, gj!

cmin764:
Also some release notes and `poetry update` (to **main** and **meta** packages) are missing. (even a _minor_ version bump if you plan to release immediately)

packages/main/src/RPA/Browser/Selenium.py
osrjv: functools.wraps effectively makes the outer look like the inner, but maybe a bit too much in this inheritance case. it also replaces attributes like `__qualname__` and `__module__`, and I'm not confident that won't have some surprising side-effects.

PR #508, opened 2022-05-16
RPA.Cloud.AWS Create Redshift Data service and STS service

### LINKED ISSUE #496: Implement `redshift-data` service in `RPA.Cloud.AWS`
Implement the API functions available in the Redshift Data API service of boto3:

- Boto3 documentation of [RedshiftDataAPIService](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/redshift-data.html)
- [AWS Redshift Data API reference](https://docs.aws.amazon.com/redshift-data/latest/APIReference/Welcome.html)
- Getting started guide for querying Redshift data: [Querying a database](https://docs.aws.amazon.com/redshift/latest/mgmt/query-databases.html)

### LINKED ISSUE #498: `RPA.Cloud.AWS` - Implement `sts` role assumption
The `RPA.Cloud.AWS` library needs to implement the `sts` service and be able to handle a request to assume a role, see the guide on [Switching to an IAM role](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-api.html).

This guide utilizes boto3 resources instead of services to execute the actions, while the entire library uses services. We will have to test if role assumption can work with services. I believe if we can reinitialize the client with the temporary credentials provided back to use from the `AssumeRole` operation when creating a new service, it should work.

blakewaud:
This PR creates two new services for RPA.Cloud.AWS: Redshift Data and STS.

* The STS service allows for the use of `Assume Role` (fixes #498). All services were updated to be able to accept the session token necessary to make the temporary credentials work.
* The Redshift Data service can be used to send SQL statements to Redshift data clusters (fixed #496).

Included in this PR is some mocked testing using the `moto` package, which has been added to the dev dependancies.


blakewaud:
I am working on release notes and any documentation updates now.

packages/aws/src/RPA/Cloud/AWS/__init__.py
```diff
@@ -116,17 +123,16 @@ def _init_client(
...
-                " to use AWS cloud service: %s" % service_name
-            )
+            auth_params = {}
+        else:
+            auth_params = {
+                "aws_access_key_id": aws_key_id,
+                "aws_secret_access_key": aws_key,
+            }
+        if session_token:
+            auth_params["aws_session_token"] = session_token
```
cmin764, line 133: Q: Just to be sure, will the session token take precedence (when read by the server) in front of the key ID & access key?

packages/aws/src/RPA/Cloud/AWS/__init__.py
```diff
@@ -100,14 +106,15 @@ def _init_client(
         aws_key: str = None,
         region: str = None,
         use_robocloud_vault: bool = False,
+        session_token: str = None,
```
cmin764, line 109:
```suggestion
        session_token: Optional[str] = None,
```
here and everywhere else.

packages/aws/src/RPA/Cloud/AWS/__init__.py
```diff
@@ -990,7 +1024,672 @@ def delete_queue(self, queue_name: str = None):
...
+        single SQL statement.
+
+        SQL statements can be parameterized with named parameters through
+        the use of the ``parameters`` argument. Parameters must be dictionaries
+        with the following two keys:
+
+        * ``name``: The name of the parameter. In the SQL statement this
+          will be referenced as ``:name``.
+        * ``value``: The value of the parameter. Amazon Redshift implicitly
+          converts to the proper data type. For more inforation, see
```
cmin764, line 1108:
```suggestion
          converts to the proper data type. For more information, see
```

packages/aws/src/RPA/Cloud/AWS/__init__.py
```diff
@@ -990,7 +1024,672 @@ def delete_queue(self, queue_name: str = None):
...
+            does not utilize a timeout directly.
+
+        """  # noqa: W605, E501
+        client = self._get_client_for_service("redshift-data")
+        run_token = self._submit_statement(
+            client, sql, parameters, statement_name, with_event
+        )
+        self.logger.info(
+            f"'{run_token['Id']}' SQL statement execution on Redshift started"
+            f"{' with name ' + statement_name if statement_name else ''}:\n{sql}"
```
cmin764, line 1168: I find this `f""` formatting and its curly braces a little bit tricky here. What's the supposed outcome of formatting the `{ with name ...}` part? (I find hard to understand the intended behaviour)

packages/aws/src/RPA/Cloud/AWS/__init__.py
```diff
@@ -990,7 +1024,672 @@ def delete_queue(self, queue_name: str = None):
...
+        will wait for results. See \`Execute Redshift Statement\` for
+        additional information.
+
+        If the statement has tabular results, this keyword returns them
+        as a table from ``RPA.Tables`` if that library is available, or
+        as a list of dictionaries if not. If the statement does not have
+        tabular results, it will return the number of rows affected.
+
+        :param statement_id: The statement id to use to retreive results.
+        :param timeout: An integar used to calculate the maximum wait.
```
cmin764, line 1240:
```suggestion
        :param timeout: An integer used to calculate the maximum wait.
```

packages/aws/src/RPA/Cloud/AWS/__init__.py
```diff
@@ -990,7 +1024,672 @@ def delete_queue(self, queue_name: str = None):
...
+            output = float(tagged_union["doubleValue"])
+        elif tagged_union.get("isNull"):
+            output = None if tagged_union["isNull"] else "UNKNOWN_DATA_MEMBER"
+        elif tagged_union.get("longValue"):
+            output = int(tagged_union["longValue"])
+        elif tagged_union.get("stringValue"):
+            output = str(tagged_union["stringValue"])
+        else:
+            output = "UNKNOWN_DATA_MEMBER"
+        return output
```
cmin764, line 1334: Might be more maintainable to have a constant dict here with `{"blobValue": bytes, ...}` and iterate over it to get the values from `tagged_union` associated with their casting types.

packages/aws/src/RPA/Cloud/AWS/__init__.py
```diff
@@ -990,7 +1024,672 @@ def delete_queue(self, queue_name: str = None):
...
+        )
+
+    @aws_dependency_required
+    def assume_role(
+        self,
+        role_arn: str,
+        role_session_name: str,
+        policy_arns: Optional[List[Dict]] = None,
+        policy: Optional[str] = None,
+        duration: Optional[int] = 900,
```
cmin764, line 1593:
```suggestion
        duration: int = 900,
```
**if** shouldn't be ever set to `None` (null in the payload). Which should be fine since you're not providing the keyword in the params if the value doesn't evaluate to `True`, but what if we set a duration of `0` (which evaluates as `False` at line 1678)? -- which in this case isn't added and should be the right thing to do

packages/aws/src/RPA/Cloud/AWS/__init__.py
```diff
@@ -1126,6 +1847,8 @@ def __init__(self, region: str = DEFAULT_REGION, robocloud_vault_name: str = Non
         ServiceTextract.__init__(self)
         ServiceComprehend.__init__(self)
         ServiceSQS.__init__(self)
+        ServiceRedshiftData.__init__(self)
```
cmin764, line 1850: Peculiar way of initialising the super classes, where a `super().__init__()` will take care of all the upper inits. (if we keep the idiom in all super-classes `__init__` methods) -- might be a good time for some opportunistic refactoring here.

docs/source/releasenotes.rst
```diff
@@ -20,6 +20,16 @@ Release notes
 - Library **RPA.Browser.Selenium** (:pr:`502`): Automatically add URL scheme when
   navigating, such as `https` (default) or `http`. This functionality is controlled
   with the keyword `Set default URL scheme`.
+- Library **RPA.Cloud.AWS**:
```
cmin764, line 23: This moved now since we just released the rest of the notes.

packages/aws/tests/python/test_redshift_data.py
```diff
@@ -0,0 +1,149 @@
...
+    )
+    # data = boto3.client("redshift-data", region_name=DEFAULT_REGION)
+    # test_resp = data.execute_statement(
+    #     Sql=SELECT_ALL_TABLES_SQL,
+    #     ClusterIdentifier=TEST_CLUSTER,
+    #     DbUser=TEST_USER,
+    #     Database="dev",
+    # )
+    # print("\n\nRESPONSE:")
+    # print(test_resp)
```
cmin764, line 107: Are these commented code blocks intended to be uncommented and having them working afterwards? Otherwise not sure about the value it brings to have them here, but I suppose this is still WIP somehow.

packages/main/README.rst
```diff
@@ -398,22 +399,52 @@ First steps to start developing:
...
+   delete the ``RPA`` folder and associated ``distinfo`` folders in the metapackage's **.venv** directory. The doc gen
+   command will reinstall the local packages.
 
    - ``poetry update``
    - ``make docs``
-   - open ``docs/build/html/index.html`` with the browser to view the changes
+   - open ``docs/build/html/index.html`` with the browser to view the changes or execute ``make local`` and navigate to
+     ``localhost:8000`` to view docs as a live local webpage.
+
+   **NOTE:** Due to the way optional packages are referenced in the metapackage's ``pyproject.toml``, they will not install
```
cmin764, line 421: As a side note, the other way around is `poetry install`-ing with the **meta** venv activated the optional packages (like `aws`) you want available for dev given the most recent version you have locally under the repo's project root dir. (this avoids re-generating a different _poetry.lock_ file or any change in the _*.toml_ one)

cmin764 (commented):
This is a pretty big PR and I haven't allocated yet sufficient time to really test it by running stuff, but I suppose you know what you're doing and it looks pretty solid on unit & robot testing.

So good job! Left a few structural nits/improvements.

docs/source/releasenotes.rst
blakewaud: Merged master to resolve

packages/aws/src/RPA/Cloud/AWS/__init__.py
blakewaud: did a regex replace throughout to resolve

packages/aws/src/RPA/Cloud/AWS/__init__.py
blakewaud: Session token is in addition to key ID and access key when using temporary credentials, without it, auth will fail.

packages/aws/src/RPA/Cloud/AWS/__init__.py
blakewaud: Thanks for catching

packages/aws/src/RPA/Cloud/AWS/__init__.py
blakewaud: Broke the if statement to a line before to clarify behavior

packages/aws/src/RPA/Cloud/AWS/__init__.py
blakewaud: Thank you

packages/aws/src/RPA/Cloud/AWS/__init__.py
blakewaud: Good point, needed to define a minimum anyway.

packages/aws/src/RPA/Cloud/AWS/__init__.py
blakewaud: Good point, I hadn't thought about how `super()` obviously hits all of the inherited classes.

packages/aws/tests/python/test_redshift_data.py
blakewaud: Yes that was a hold over and I forgot to delete them.

packages/aws/src/RPA/Cloud/AWS/__init__.py
blakewaud: Great idea, thanks, this is much more clean

tasks.py
```diff
@@ -0,0 +1,122 @@
...
+
+GIT_ROOT = _git_root()
+PACKAGES_ROOT = GIT_ROOT / "packages"
+
+if platform.system() != "Windows":
+    ACTIVATE_PATH = GIT_ROOT / ".venv" / "bin" / "activate"
+    ACTIVATE = f"source {ACTIVATE_PATH}"
+else:
+    ACTIVATE_PATH = GIT_ROOT / ".venv" / "Scripts" / "activate"
+    ACTIVATE = f"{ACTIVATE_PATH}.bat"
```
cmin764, line 25: I believe we can remove these (and their usage) since the current meta venv context is already activated when you're issuing Poetry-based invoke commands with: `poetry run inv <cmd>` (which is the correct way to do it since _invoke_ is covered as a Poetry dep)

Is there any edge-case when we need to explicitly activate the venv first? (like a case where is not activated yet or when not running under the venv Poetry context)

cmin764: Sorry, what I meant was:
1. You run invoke commands with `poetry run inv <cmd>` in your terminal.
2. Your poetry venv gets automatically activated so it runs invoke within that environment.
3. No need to re-activate the venv in the _tasks.py_ file since the meta venv is already activated. (and any other `poetry install` of other projects happens in this already activated venv)

Usually you run once `poetry install` to create the venv then issue any other command with `poetry run ...` like:
- `poetry run inv ...`
- `poetry run python ...`
- `poetry run pip ...`
This ensures the executable runs from the venv. (and btw, on a clean env you don't usually have `inv`, you only have Python and Poetry and you start from there, that's why might be the safest to just `poetry run` everything; also you don't need to run `poetry run poetry install` because the installation part is covered in the _tasks.py_ file)

All from above (with my change) was tested on both Mac & Windows and is working just fine, that's why I was curious if is there any expected intentional behaviour from activating again explicitly the meta venv inside the _tasks.py_ file.

blakewaud: Ah, I understand now. The current readme only shows using invoke like `invoke lint`. I had not thought to execute all invoke commands via `poetry run`. That would solve the issue... perhaps that's something to be added to the readme as I hadn't thought of that at all before now.

My only concern is if I write the invoke tasks assuming everyone knows to execute via `poetry run`, that's a possibly bad assumption. Perhaps I can write an error message that can tell it was not ran via `poetry run`

blakewaud: @cmin764 I am updating the readme to indicate invoke should be ran via poetry, but I am looking at the `tasks.py` for the main package and I see the definition for the `lint` task executes a `poetry run` command inside of it. I tested it and it doesn't cause it to break... but I guess this begs the question what is expected with `invoke`. Do all devs expect to run `invoke` via `poetry run` or should we assume the other?

That implementation was written by @mikahanninen so perhaps he can weigh in on preference. I just need to know if I should assume inside an invoke task that someone started the task via `poetry run` or not.

cmin764: Understood, all these `'run ...` commands are for paranoia-ensuring that Invoke will use the executables from the venv, so you're right, it is expected to have developers using `inv <cmd>` directly, even outside of the venv. Hence, a paranoia inner venv activation won't hurt and be aligned with rest of the code, funny thing I added `invoke` into **main** [here](https://github.com/robocorp/rpaframework/pull/309/files#diff-b99f89fed4ef40ec0f6699d9ef63897336d08d867f2e6db306b64db7e01f78f7) as a monkey-see monkey-do from **windows**. Currently these **bold** packages have `invoke` listed in their requirements:
- meta
- **aws** (Mika)
- core
- dialogs
- **google** (Mika)
- **main** (me)
- **pdf** (me)
- recognition
- **windows** (Mika)

So definitely we should decide if is expected to develop with the Poetry-installed Invoke or with a standalone external one installed separately by the developer (or with both if we want to support that), then align all packages' requirements given that.

blakewaud: Based on discussion on slack, I am assuming using external `invoke` and activating the environment for all my actions. Shouldn't hurt if you did use `poetry run invoke` though.

tasks.py
```diff
@@ -101,22 +107,29 @@ def install_local(ctx, package):
     editable form instead of from PyPi. This task always resets
     the virtual environment first.
 
-    Package must exist as a subfolder module in ``\\packages``,
+    Package must exist as a sub-folder module in ``./packages``,
     see those packages' ``pyproject.toml`` for package names.
     If ran with no packages, all optional packages will be installed
-    from PyPi.
+    locally.
```
cmin764, line 113: I was hoping for this, nice!

PR #536, opened 2022-06-08
Extend work item examples

blakewaud:
Primarily this expands the example for `Release Input Work Item`.

Secondarily this PR includes a modification to `pyproject.toml` and adds the attribute `develop = true` to the `rpaframework` dependency in the top level meta project. This allows documents to be rebuilt without deleting the installed version of `rpaframework` every time. For this reason, I am requesting a review as this is otherwise a documentation only change.


packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -1240,25 +1291,33 @@ def get_work_item_payload(self):
...
 
     @keyword
     def set_work_item_payload(self, payload):
-        """Set the full JSON payload for a work item.
+        r"""Set the full JSON payload for a work item.
 
         :param payload: Content of payload, must be JSON-serializable
 
         **NOTE**: Most use cases should prefer higher-level keywords.
+        Using this keyword may cause errors when getting the payload via
```
cmin764, line 1308: Very good point here! (as I've seen the behaviour, but not sure exactly what other kind of payload you may have, like a list of dicts or?)

pyproject.toml
```diff
@@ -2,13 +2,15 @@
...
+authors = ["RPA Framework <rpafw@robocorp.com>"]
 
 [tool.poetry.dependencies]
 python = "^3.7"
-rpaframework = { path = "packages/main", extras = ["cv", "playwright", "aws"] }
+rpaframework = { path = "packages/main", extras = [
+	"cv",
+	"playwright",
+	"aws",
+], develop = true }
```
cmin764, line 13: Rebuilt my meta venv with this branch and it works like it should!
```
(.venv) cmin@robocorp-mac rpaframework % python -c "import RPA.Robocorp.WorkItems as w; print(w)"
<module 'RPA.Robocorp.WorkItems' from '/Users/cmin/Repos/rpaframework/packages/main/src/RPA/Robocorp/WorkItems.py'>
```

Just to keep in mind that only `rpaframework` gets installed for development in this meta venv, the rest of the deps (including our `extras`) are installed normally.

cmin764 (approved):
LGTM!

A couple of nitpick comments and questions left so we can learn more together about it.

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -852,6 +854,17 @@ class WorkItems:
     through the "RPA_OUTPUT_WORKITEM_PATH" env var a different path and name for this
     file.
 
+    **Mocking with Robocorp Code VSCode Extension**
```
mikahanninen, line 857: Mocking => Developing

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -852,6 +854,17 @@ class WorkItems:
...
+
+    If you are developing in VSCode, the `Robocorp Code extension`_, you can
+    utilize the built in mocking features described in the
```
mikahanninen, line 860: "mocking features" => "development features" or just "features"

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -1081,24 +1094,37 @@ def _end_suite(self, data, result):
...
     @keyword
     def set_current_work_item(self, item: WorkItem):
-        """Set the currently active work item.
+        r"""Set the currently active work item.
 
         The current work item is used as the target by other keywords
         in this library.
 
-        Keywords ``Get Input Work Item`` and ``Create Output Work Item``
+        Keywords \`Get Input Work Item\` and \`Create Output Work Item\`
```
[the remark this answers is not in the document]
mikahanninen: Yeah I would not use raw docstings. Was there something wrong with double ``keyword`` ?

mikahanninen (changes requested):

packages/main/src/RPA/Robocorp/WorkItems.py
blakewaud: So I used "mock" in this instance as it is essentially a continuation and alternate way of performing was Ossi originally wrote just above it titled "Development and mocking" as this is the part were a robot developer needs to mock a work item so they can test their robot. Should we change Ossi's documentation as well to avoid the word mock entirely? I am not sure as that may lose the meaning of the doc section.

cmin764: No strong opinion here, but _Mocking_ in this scenario might be clearer since you're mocking a feature covered by the Cloud. (as in the same way you can use the CR Vault or the local _mocked_ Vault right from VSCode)

But at the same time that's not exactly mocking, it is actually local development and cloud environment simulation, so I'd go with Mika's suggestion if Ossi agrees as well to use the same term everywhere. (as I don't have a strong-opinion myself with this one)

blakewaud: Based on slack discussion, switching all references of "mock" to "local development" or "cloud simulation".

packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -852,6 +854,17 @@ class WorkItems:
     through the "RPA_OUTPUT_WORKITEM_PATH" env var a different path and name for this
     file.
 
+    **Mocking with Robocorp Code VSCode Extension**
+
+    If you are developing in VSCode, the `Robocorp Code extension`_, you can
```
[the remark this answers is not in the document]
blakewaud: Your first suggestion is what I meant, thanks.

packages/main/src/RPA/Robocorp/WorkItems.py
blakewaud: As above, removing "mock".

packages/main/src/RPA/Robocorp/WorkItems.py
blakewaud: Or just a list, yes. Payload needs to have a `dict` as it's only contents or the `Get work item variable` keyword will fail because that just pulls a key off the `dict` attached to `Payload`.

PR #539, opened 2022-06-09
Work Items: Auto-release input work item on failure

### LINKED ISSUE #538: Failure during work item processing should report correct exception type
If some unexpected exception during work item processing is raised and the execution fails, Control Room will report the work item as failed with the type `ORCHESTRATOR`, which is technically incorrect. It's should be `APPLICATION`, but the orchestrator has no way of knowing that.

Our library should have a catch-all when the execution fails and it should release the work item with the correct type. It's also an opportunity to propagate the error message from the exception.

osrjv:
Closes #538 

### ToDo
- [x] Unit ~& robot~ test
- [x] Real Control Room test with: setup/teardown suite failure & normal task failure


packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -1070,28 +1070,50 @@ def _parse_work_item_from_email(self):
             # pylint: disable=undefined-loop-variable
             self.set_work_item_variable(keys[0], parsed_email)
 
-    def _start_suite(self, data, result):
+    def _start_suite(self, *_):
```
bogdancondurache, line 1073:
```suggestion
    def _start_suite(self, *args, **kwargs):
```

I guess this would be the Python way of doing it

PR #541, opened 2022-06-09
Deprecate Lab references

cmin764:
An attempt on removing the Lab related references.

### DIFF
--- a/packages/main/src/RPA/Desktop/__init__.py
+++ b/packages/main/src/RPA/Desktop/__init__.py
@@ -154,9 +154,9 @@ class Desktop(DynamicCore):
     UI changes. Robot Framework scripts can then only contain a reference
     to a stored locator by name.
 
-    The main way to create named locators is with `Robocorp Lab`_.
+    The main way to create named locators is with `VSCode`_.
 
-    .. _Robocorp Lab: https://robocorp.com/docs/product-manuals/robocorp-lab/robocorp-lab-overview
+    .. _VSCode: https://robocorp.com/docs/developer-tools/visual-studio-code/overview
 
     **Keyboard and mouse**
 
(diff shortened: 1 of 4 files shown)


packages/main/src/RPA/Desktop/__init__.py
```diff
@@ -154,9 +154,9 @@ class Desktop(DynamicCore):
     UI changes. Robot Framework scripts can then only contain a reference
     to a stored locator by name.
 
-    The main way to create named locators is with `Robocorp Lab`_.
+    The main way to create named locators is with `VSCode`_.
 
-    .. _Robocorp Lab: https://robocorp.com/docs/product-manuals/robocorp-lab/robocorp-lab-overview
+    .. _VSCode: https://robocorp.com/docs/developer-tools/visual-studio-code/overview
```
cmin764, line 159: Shall we replace this vague reference with this article instead: https://robocorp.com/docs/development-guide/desktop#image-template-matching-based-automation

bogdancondurache: Just my personal perspective: I always love it when URL's point my to very helpful and specific things instead to the generic one from where I have to find the specific one :) So I think going with the article would be better. Which article, I don't know... This one also seems very helpful https://robocorp.com/docs/development-guide/desktop/how-to-find-user-interface-elements-using-locators-and-keyboard-shortcuts-in-windows-applications
Actually might be even more helpful as it's more practical. But I guess any of them works

kariharju: We need to re-write the locator docs to a format that is separated from a single tool.

mikahanninen: I am not sure of the deprecation messages like  `Robocorp Lab (deprecated)`

@kariharju shouldn't we just remove Lab references cleanly ?

PR #569, opened 2022-07-01
`RPA.Email.Exchange` OAuth2 support

### LINKED ISSUE #567: `RPA.Email.Exchange` Implement Oauth2 flow as basic auth is being deprecated
Per the top of the [Exchange Web Services API Overview page](https://docs.microsoft.com/en-us/exchange/client-developer/exchange-web-services/get-started-with-ews-managed-api-client-applications), Microsoft is deprecating basic authentication by October 2022. We need to implement the Oauth2 flow in its place. The underlying `credentials.py` appears to already have a class for Oauth2.

cmin764:
Closes #567 
Working Portal [robot](https://github.com/robocorp/example-oauth-email/blob/feature/microsoft-oauth2/tasks.robot#L73-L79)

### ToDo
- [x] Optionally add a robot test (for displaying how the flow works)
  - note that the unit-test is already skipped even for the basic auth case

### DIFF
--- a/packages/main/src/RPA/Email/Exchange.py
+++ b/packages/main/src/RPA/Email/Exchange.py
@@ -7,7 +7,7 @@ import time
 from email import policy  # pylint: disable=no-name-in-module
 from multiprocessing import AuthenticationError
 from pathlib import Path
-from typing import Any, List, Optional, Union
+from typing import Any, Callable, List, Optional, Union
 
 import pytz
 from exchangelib import (
@@ -57,6 +57,18 @@ EMAIL_CRITERIA_KEYS = {
 
 class NoRecipientsError(ValueError):
     """Raised when email to be sent does not have any recipients, cc or bcc addresses."""  # noqa: E501
+
+
+class OAuth2Creds(OAuth2AuthorizationCodeCredentials):
+
+    """OAuth2 auth code flow credentials wrapper supporting token state on refresh."""
+
+    def __init__(self, *args, on_token_refresh: Callable, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._on_token_refresh = on_token_refresh
+
+    def on_token_auto_refreshed(self, token: OAuth2Token):
+        self._on_token_refresh(token)
 
 
 class Exchange:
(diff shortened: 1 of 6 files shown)


packages/main/src/RPA/Email/Exchange.py
```diff
@@ -223,13 +255,37 @@ def authorize(
         kwargs["primary_smtp_address"] = (
             primary_smtp_address if primary_smtp_address else username
         )
-        self.credentials = Credentials(username, password)
+        if is_oauth:
+            self.credentials = OAuth2Creds(
+                on_token_refresh=self.on_token_refresh,
```
blakewaud, line 260: I'm guessing there's some code I'm not following in the `oauthlib` library, but from what I see, we pass in a function that only logs to override the auto refresh method. Assuming I am wrong about this, you can ignore my comment and I'd appreciate learning how this is actually working.

blakewaud (approved):
LGTM! One question on the refresh functionality likely for my own understanding.

packages/main/src/RPA/Email/Exchange.py
cmin764: You're right 100% and at first glance this seems useless, but it actually adds support for customizing token refresh behaviour with ease in your robot, like in this [example](https://github.com/robocorp/example-oauth-email/pull/1/files#diff-b93aca6dec83011aab894b95b6deaaf19b48d5870212c8366f1e0e66acd3cc80R29-R33):
```python
class ExtendedExchange(Exchange):

    def __init__(self, *args, secret_name, **kwargs):
        super().__init__(*args, **kwargs)
        self._secret_name = secret_name

    def on_token_refresh(self, token):
        super().on_token_refresh(token)
        secret = vault.get_secret(self._secret_name)
        secret["token"] = dict(token)
        vault.set_secret(secret)
```
which saves the token in the vault on refresh so the new one can be picked up next time the robot decides to run again to authorize and send an e-mail.

blakewaud: That's great... and now I'm thinking your comment should be added to the documentation as an advanced use case somehow! 😄

cmin764: Yes, not yet in the docs but exists in the docstring of the `on_token_refresh` method and most of the instructions will go in the Portal [bot](https://robocorp.com/portal/robot/robocorp/example-oauth-email) which show-cases the extension of `Exchange`. I need at least one customer dealing with OAuth2 flow on `RPA.Email.Exchange` through the library before committing any more focus into this. (as I bet most of them will still prefer or be able to use just "App Passwords")

PR #602, opened 2022-08-11
Upgrade to Selenium 4

mikahanninen:
TODO
- [x] core release with selenium 4.4.0 included
- [ ] main release using updated core

Issue #494, opened 2022-05-09
Migrate from Selenium 3 to Selenium 4


mikahanninen:
### Affected rpaframework "parts"

- `rpaframework-core` package ([webdriver-manager](https://pypi.org/project/webdriver-manager/) update)
- `rpaframework` package and especially `RPA.Browser.Selenium` library
- dependency `robotframework-seleniumlibrary` needs to be upgraded to minimum version of **6.0.0**
- dependency `robotframework-seleniumtestability` needs to be upgraded to minimum version of **2.0.0**
- example bots

### Other areas to be updated

- tests
- documentation

### Links

- [Article on the migration](https://www.browserstack.com/guide/upgrade-from-selenium-3-to-selenium-4)
- [List of Selenium API deprecations](https://www.selenium.dev/selenium/docs/api/java/deprecated-list.html)
- Battle testing [repo](https://github.com/mikahanninen/selenium4-battleground)

### TODO / problem areas

- [x] `core` package release with updated dependencies. (#607 )
- [x] Caching is tackled now by the `webdriver-manager`, so remove `webdriver.cache(browser)` usage from `Selenium`.
- [x] Custom `options` argument support in the OAB keyword. (#385 #258)
- [x] `DesiredCapabilities` got replaced by `Options`. (so our lib extensions should)
- [x] Go through all the [deprecated classes](https://www.selenium.dev/selenium/docs/api/java/deprecated-list.html) and cross-check with our internal usage. (library, examples, lib extensions)
- [x] Update release notes.
- [ ] `main` package release including updated `core` + changes to `RPA.Browser.Selenium`.
- [ ] The [Example Web Store Processor](https://github.com/robocorp/example-web-store-order-processor) (Portal) bot is using `find_element_by_class_name` on the returned `WebElement` object and that no longer exists. Need to be replaced with generic `find_element` and pass locator strategy as a first parameter.
- [x] Ask people for battle testing this before official release. (DevPI wheel and robot example)

###### Selenium 3
`${add_to_cart_button} =    Set Variable    ${product.find_element_by_class_name("btn_primary")}`

###### Selenium 4
`${add_to_cart_button} =    Set Variable    ${product.find_element("class name", "btn_primary")}`

PR #606, opened 2022-08-23
Use `webdriver-manager` with `core` for Selenium 4

cmin764:
Potentially breaking change making use of another web driver manager called [webdriver-manager](https://pypi.org/project/webdriver-manager/).

### ToDo
- [ ] ~Unit-tests~ (not making sense at this stage)
- [x] Real end-to-end tests with [Chrome, Firefox, Opera, Edge] x [Windows, Linux, Mac]

Dependency of #602 
Relates to #494


packages/core/src/RPA/core/webdriver.py
```diff
@@ -21,94 +27,60 @@
...
 }
+AVAILABLE_DRIVERS = {
+    # Driver names taken from `webdrivermanager` and adapted to `webdriver_manager`.
+    "chrome": ChromeDriverManager,
+    "firefox": GeckoDriverManager,
+    "gecko": GeckoDriverManager,
+    "mozilla": GeckoDriverManager,
+    "opera": OperaDriverManager,
+    # NOTE: There's no specific `EdgeDriverManager` with this manager.
+    "edge": EdgeChromiumDriverManager,
```
osrjv, line 38: Not sure what is the best approach here. Edge referred to the old edge which isn't chromium-based. We're dropping the support here, but I'm a bit on the fence if we should just remove the alias or point towards chromium edge. New users would probably expect to get chromium edge instead of the old one, maybe.

cmin764: Thank you for the observation!

Also an interesting question is if the `selenium.webdriver.Edge` at least works with the chromium-edge driver. (haven't tested)

And there was a bug here proving that nobody used the Chromium flavour of Edge because we had `edgechromium` expecting for a browser name "EdgeChromium" which will fail the `getattr` on the `webdriver` module because we actually have the "ChromiumEdge" available (reversed lexical atoms) as seen [here](https://github.com/SeleniumHQ/selenium/blob/selenium-3.150.0/py/selenium/webdriver/__init__.py#L26). (in both Selenium 3 and 4)
So all our Edge users we'll be redirected to ChromiumEdge until this upstream [Issue](https://github.com/SergeyPirogov/webdriver_manager/issues/431) gets sorted out.

osrjv: If I remember correctly, non-chromium Edge requires a completely different webdriver, and it had to be installed via the Windows add/remove features menu.

osrjv: Possibly why it's not supported by webdriver-manager.

cmin764: Since in Selenium 4 `Edge` became the same as `ChromiumEdge` I feel is safe to keep the same driver manager for both and assume than `edge` is just an alias for `chromiumedge` which is the only one currently supported now.

cmin764:
Tested current code with the following combinations so far:

```robotframework
Open Available Browser    https://robocorp.com    browser_selection=${browser}
```

- Mac: Chrome, Firefox, Edge (chromium)
- Windows: Chrome, Firefox, Edge (fails, browser passed as `None`), ChromiumEdge (unsupported browser -- need to check why)
- Linux: _not yet tested on_

### Remarks
- Opera was dropped in Selenium 4
- Edge got replaced by ChromiumEdge and `edge` & `chromiumedge` are the same now
- Caching is tackled automatically by the `webdriver_manager` package

mikahanninen (approved):

cmin764:
Tested with [robot](https://github.com/cmin764/robots/blob/main/browser-auto/tasks.robot#L121-L125) using Chrome, Firefox, ChromiumEdge on both Mac & Windows and is working just fine.

On Linux I get Chrome working fine, but with Firefox:
```console
BrowserNotFoundError: Failed to start a browser:
- Firefox: Message: Expected browser binary location, but unable to find binary in default location, no 'moz:firefoxOptions.binary' capability provided, and no binary flag set on the command line
```

Screenshot 2022-08-26 at 14 00 12


Need to fix Firefox first before merging this.

And with ChromiumEdge we might be able to run in headless mode as well. (if not, using it on a non-UI Linux container won't work obviously)

PR #631, opened 2022-09-16
Updated RPA.Salesforce to better handle the `sandbox=${TRUE}` flag

TimothyLopas:
Documentation around the different ways to set the `sandbox` flag has been updated as well as a new keyword `Set Domain` to make setting a domain other then 'login' easier.


packages/main/src/RPA/Salesforce.py
```diff
@@ -176,6 +217,12 @@ def _require_authentication(self) -> None:
         if self.sf is None:
             raise SalesforceAuthenticationError("Authentication is not completed")
 
+    def _require_no_session(self) -> None:
+        if self.session_id is not None or self.instance is not None:
```
cmin764, line 221: A simpler:
```suggestion
        if self.session_id or self.instance:
```
might work the same.

packages/main/src/RPA/Salesforce.py
```diff
@@ -176,6 +217,12 @@ def _require_authentication(self) -> None:
         if self.sf is None:
             raise SalesforceAuthenticationError("Authentication is not completed")
 
+    def _require_no_session(self) -> None:
```
cmin764, line 220:
```suggestion
    def _raise_on_session(self) -> None:
```
if you like it more.

cmin764 (approved):
LGTM! (left a few nitpicks)

@mikahanninen Do you have any idea on how to create a Salesforce **sandbox** to test with? [This](https://github.com/cmin764/robots/blob/main/enterprise/salesforce.robot) is how I've been testing it, but on a real instance, not a sandbox one.

@TimothyLopas Maybe this [article](https://www.salesforceben.com/salesforce-sandbox/#:~:text=the%20new%20sandbox.-,What%20Is%20a%20Salesforce%20Sandbox%3F,Accounts%2C%20Contacts%2C%20and%20Leads.) will help with the **sandbox** creation and testing, give it a try.

TimothyLopas:
> 

@cmin764: I read through the article and it appears that I cannot setup a sandbox environment because my instance is a free developers instance. A paid license appears to be necessary to setup a sandbox environment. 

Found this in a different article: "Note: If you are using a free Developer Salesforce org, it already is a sandbox so the sandbox functionality described here does not apply, you cannot create a sandbox within a sandbox environment."

packages/main/src/RPA/Salesforce.py
```diff
@@ -163,10 +201,13 @@ class Salesforce:
...
 
-    def __init__(self, sandbox: bool = False) -> None:
+    def __init__(self, sandbox: bool = False, domain: str = "login") -> None:
         self.logger = logging.getLogger(__name__)
         self.sf = None
-        self.domain = "test" if sandbox else "login"
+        if sandbox:
+            self.set_domain("sandbox")
+        else:
+            self.set_domain(domain)
```
cmin764, line 210:
```suggestion
        self.set_domain("sandbox" if sandbox else domain)
```

packages/main/src/RPA/Salesforce.py
```diff
@@ -163,10 +201,13 @@ class Salesforce:
 
     account = {"Name": None, "Id": None}
 
-    def __init__(self, sandbox: bool = False) -> None:
+    def __init__(self, sandbox: bool = False, domain: str = "login") -> None:
         self.logger = logging.getLogger(__name__)
         self.sf = None
-        self.domain = "test" if sandbox else "login"
```
cmin764, line 169: Optional nit:
`if sandbox and domain != "login": raise SalesforceDomainChangeError("Can't use a custom domain with sandbox on")`

Feel free to change the error type/message and use a constant like `DEFAULT_DOMAIN = "login"`. Or not apply this suggestion if you feel that's breaking or too intrusive.

cmin764 (approved):
LGTM!

- Left 2 more comments. (like what do we do if we also set custom domain and sandbox; a refactor nitpick)
- Optionally you can add a quick robot/unit test to check the following behaviour:
   - `Set Domain` works when there's both no session & instance (check if domain changes)
   - Passing sandbox or custom domain during init -> check how domain changes

packages/main/src/RPA/Salesforce.py
TimothyLopas: Based on the conditional in the __init__ function I believe that setting sandbox = true will override the custom domain. So this may be unnecessary. Correct me if I missed something.

cmin764: Correct! But what if the user provides both by mistake and doesn't know what he/she's doing? We can help him/her with an error :D.

Since you merged and this is not as important anymore, we can leave it flexible like it is.

PR #650, opened 2022-10-03
`RPA.DocumentAI` Intelligent Document Processing generic keywords

### LINKED ISSUE #557: New library: `RPA.DocumentAI` - generic keywords to be used with Google, Base64 and Nanonets
Set of generic keywords to be used with the latest available engines for processing documents: Google DocumentAI, Base64, Nanonets.

Place for the library: `RPA.DocumentAI`.
Engine names: `google`, `base64ai`, `nanonets`.

### Keywords

- `Init Engine    <name>  [    <secret>  |  vault={"store_name": "secret_key"}    <params>]`: Prepares the `<name>` engine for being able to process documents. At least one secret is required (passed directly or enabled through Vault) and based on the selected engine, you need to provide other parameters as well, like the e-mail address, region etc.
   - `<secret>`: Can be a string API key. If a file path is provided instead, we'll use the content inside of it as for the value. If such value is passed to the `google` engine, then it should be a file path pointing to either a service account or token JSON file. (and based on its content, we detect the type of it: service_account / token)
   - `<vault>`: The Vault stored secret location as `key: value` directive, where `key` is the store name and `value` is the secret key. If provided, we pick up the secret from here instead. And in order to support `google` too, we'll be reusing the `get_secret_from_robocorp_vault` context method to operate internally on the Vault easily.
   - If no secret is provided at all, `Init Engine` acts like a `Switch Engine`, which will simply switch the context to another engine to be used. (assuming that the one we're switching on is already initialized with the necessary secret to function, otherwise it errors during the call of `Predict`)
- `Predict    <file-path/URL>   [    <model>    <params>]`: Auto-detects if the input source is coming from a local file address or remote URL one. Then it uses the `<model>` (model ID or type or even processor ID with `google`) to instruct the prediction algorithm with. Optional `<params>` are supported, like with `google` we need: project ID, region and optionally a mime type.
- `${result} =    Get Result`: Uses the internal stored context result obtained by `Predict` to further create a human-friendly dictionary output out of it. The user is free to understand and interpret such JSON-like output as it wants. (we won't enforce a standard structure)

----

Later on we can support OCR related engines as well by using the very same flow described above.

Currently available OCR libraries (aka adapters):
- `RPA.Cloud.AWS`, the **Textract** service (rpaframework-aws)
- `RPA.Cloud.Azure`, the **Vision** service (rpaframework)
- `RPA.Cloud.Google`, the **Vision** service (rpaframework-google)
- `RPA.Desktop` (rpaframework + rpaframework-recognition)
- `CaptureFast` (robotframework-capturefast)

cmin764:
Closes #557 

### ToDo
- [x] API setup
- [x] Complete keywords with logic, documentation and fix TODOs
- [x] Unit & robot tests
- [ ] Basic test example ran in CI with DevPI pre-release build
- [x] Major bump release & notes (breaking import changes)

### Next
- Library docs site section entry (+ relinks of `RPA.DocumentAI.Base64AI` and `RPA.DocumentAI.Nanonets`)
- Portal [example](https://github.com/robocorp/example-document-ai) repo registration
- 10 Nov: **AWS** and **Azure** support

### DIFF
--- a/packages/main/src/RPA/DocumentAI.py
+++ b/packages/main/src/RPA/DocumentAI.py
@@ -20,12 +20,10 @@ from RPA.JSON import JSONType
 from RPA.Robocorp.Vault import Vault
 
 
-# FIXME: Type annotations globally.
-
 lib_vault = Vault()
 
 PathType = Union[Path, str]
-SecretType = Optional[Union[str, Path]]
+SecretType = Optional[Union[str, Path, Tuple, List, Dict]]
 
 
 class EngineName(Enum):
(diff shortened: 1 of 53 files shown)


mikahanninen:
I think we should support also AWS Textract as we already have that implemented.

cmin764:
> I think we should support also AWS Textract as we already have that implemented.

Sounds reasonable. Even the `Vision` one from **Azure** I believe should be supported as well, but do you mind if for this PR (to keep it smaller), we'll go on with support for the listed engines only: **Google**, **Nanonets**, **Base64**?

Then once the Portal example is ready, I can create another PR adding support for **AWS** and **Azure** as well. (including Portal example update)

packages/main/src/RPA/DocumentAI.py
```diff
@@ -0,0 +1,220 @@
...
+        elif not (secret or vault):
+            logging.warning("No `secret` or `vault` provided, relying on env vars.")
+
+        init_map = {
+            # Google library needs to be Vault aware too due to its internal way of
+            #  handling secrets from there.
+            EngineName.GOOGLE: functools.partial(self._init_google, vault=vault),
+            # Rest of the engines have a normalized way of picking up secrets either
+            #  provided directly (as string or file path) either through Vault (
+            #  dictionary of parameters).
```
cmin764, line 181: Reword this as there's a comma separated arbitrary number of values passed to the authorization keyword now.

packages/main/src/RPA/DocumentAI.py
```diff
@@ -0,0 +1,220 @@
...
+            lib_kwargs.update(
+                {
+                    "vault_name": name,
+                    "vault_secret_key": key,
+                }
+            )
+            init_kwargs["use_robocorp_vault"] = True
+        elif secret_value:
+            # Vault not used, therefore the provided secret is a file path pointing to
+            #  a service account or token JSON file.
```
cmin764, line 111: Ensure the `secret_value` is a file path. (defensive assertion)

cmin764:
Hei @mikahanninen, do you think we should include this new library in its own AI/OCR related package, like _DocumentAI/DocumentAI.py_ and inside the very same package, we also move _Base64.py_ and _Nanonets.py_ modules as well?

Then we'll be importing like so:
```
RPA.DocumentAI  # generic keywords
RPA.DocumentAI.Base64
RPA.DocumentAI.Nanonets
```

What do you think?

mikahanninen:
> Hei @mikahanninen, do you think we should include this new library in its own AI/OCR related package, like _DocumentAI/DocumentAI.py_ and inside the very same package, we also move _Base64.py_ and _Nanonets.py_ modules as well?
> 
> Then we'll be importing like so:
> 
> ```
> RPA.DocumentAI  # generic keywords
> RPA.DocumentAI.Base64
> RPA.DocumentAI.Nanonets
> ```
> 
> What do you think?

They can remain in the `main `package as these new libraries do not require any new dependencies which would increase package size.

cmin764:
Process [DocumentAI: Work Items](https://cloud.robocorp.com/orgrobocorp/robocorp/processes/91daa4fb-2557-41c7-9040-de790abc6629) ran successfully by e-mail trigger with the following content:

Robot [Task](https://github.com/robocorp/example-document-ai/blob/master/tasks.robot#L109-L110) (Step)

```JSON
{
  "base64ai": {
    "auth": {
      "vault": "document_ai:base64ai"
    },
    "predict": {
      "model": "finance/invoice"
    }
  },
  "nanonets": {
    "auth": {
      "vault": "document_ai:nanonets"
    },
    "predict": {
      "model": "858e4b37-6679-4552-9481-d5497dfc0b4a"
    }
  }
}
```
invoice

PR #655, opened 2022-10-06
Ensure there's no empty string in error message

cmin764:


packages/main/src/RPA/Robocorp/WorkItems.py
```diff
@@ -1092,7 +1092,7 @@ def _release_on_failure(self, attributes):
         if attributes["status"] != "FAIL":
             return
 
-        message = attributes["message"]
+        message = attributes["message"] or None
```
cmin764, line 1095: In case of empty string (`""`) this `message` will be `None` and automatically stripped from the `exception` body later on in `RobocorpAdapter.release_input` right before the API request.

PR #656, opened 2022-10-06
New keywords for the `RPA.SAP` library

mikahanninen:
- `Get Cell Value` - fixed this based on the [PR ](https://github.com/frankvanderkuur/robotframework-sapguilibrary/pull/16) on the **SapGuiLibrary** repository which has not been released
- `Get Statusbar Type` - new keyword to read message types in the status bar
- `Focus and Click` - this is a helper keyword for focusing on the element before click
- `Focus and Input Text` - this is a helper keyword for focusing on the element before inputting text
- `Generic Input Text` and `Generic Input Password` - original **SapGuiLibrary** has prevalidation on element **type**, these keywords set text if it is possible for the element regardless of the type
- `Get Element Type of Object` - this keyword is given the element instead of `element_id`, I use this in some added keywords to reduce amount `self.session.findById()` calls
- `Generic Click Element` - like the above Generic keyword this keyword does not care about the element **type**, and can be used to `select` or `press` element if element has that action available

### DIFF
--- a/docs/source/releasenotes.rst
+++ b/docs/source/releasenotes.rst
@@ -9,6 +9,7 @@ Release notes
   to ``Send Message`` / ``Send Email`` keywords. Will save the email instead of sending.
 - Library **RPA.Database** (:pr:`667`): Add SSL support for MySQL modules (``pymysql`` and
   ``mysql.connector``). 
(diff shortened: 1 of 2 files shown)


mikahanninen:
Currently under test by a client.

PR #659, opened 2022-10-10
Add error when publishing to PyPI off `master` branch

blakewaud:
Also create new customer `errors.py` file for local package error types.

If you attempt to `invoke build.publish` when your branch is not `master`, it will fail and return an error. The `master` branch can be configured with an invoke configuration if needed, but defaults to `master`.


mikahanninen (approved):

cmin764 (approved):
LGTM!

PR #682, opened 2022-11-01
New library `RPA.Assistant`

xylix:
Add the newly enhanced type of Dialogs with `RPA.Assistant`.

### done
- [x] Move TODOs/FIXMEs into a the new backlog https://github.com/robocorp/rpaframework/projects/5?add_cards_query=is:open so they won't be forgotten. (and we know what to improve next)
- [x] Solve relevant code review suggestions. (mainly from @osrjv )

### ToDo
- [x] Transfer the Python test into a skippable Robot test. (so it won't block CI given the open dialog window meant for testing until is manually closed)
- [x] Fix all linting issues, so `inv code.lint` passes and fulfills the CI gate.

(moved open ones to -> https://github.com/robocorp/rpaframework/projects/5?add_cards_query=is:open )

### Next
- Release package to PyPI.
- Portal example showcasing some of its usage.


xylix:
Screenshot 2022-11-03 at 17 08 08

Starting to have decent amount of UI elements completed. Still missing following methods:

```python3
assistant.clear_elements()
assistant.add_file_input()
assistant.add_drop_down()
assistant.add_date_input()
assistant.add_radio_buttons()
assistant.add_dialog_next_page_button()
assistant.add_submit_buttons()

assistant.show_dialog()
assistant.wait_dialog()
assistant.wait_all_dialogs()
assistant.close_dialog()
assistant.close_all_dialogs()
assistant.wait_dialogs_as_completed()
```

packages/assistant/src/RPA/Assistant/date_picker.py
```diff
@@ -0,0 +1,87 @@
+from flet import Dropdown, UserControl, Container, Row, Icon, PopupMenuButton, PopupMenuItem, Text
+from flet.dropdown import Option
+from flet.icons import CALENDAR_MONTH, ARROW_BACK_ROUNDED, ARROW_FORWARD_ROUNDED
+from flet import MainAxisAlignment
+from datetime import datetime, date
+from calendar import monthrange
```
cmin764, line 6: Nit: `isort` tool can be used here to sort deps into standard, 3rd-party and 1st-party groups.

packages/assistant/src/RPA/Assistant/date_picker.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class DatePicker(UserControl):
```
cmin764, line 9: Nit: A docstring can be placed here in case Pylint complains during the `inv code.lint` run.

packages/assistant/src/RPA/Assistant/date_picker.py
```diff
@@ -0,0 +1,87 @@
...
+        self.header_element[1].value = self.selected_year
+        self.update()
+
+    def _year_decrement(self, e):
+        self.selected_year -= 1
+        self.header_element[1].value = self.selected_year
+        self.update()
+    
+    def _select_month(self, e):
+        months = (("Jan", "Feb", "Mar", "Apr"), ("May", "Jun", "Jul", "Aug"), ("Sep", "Oct", "Nov", "Dec"))
```
cmin764, line 22: Q: Why months are grouped like in 3 kind of quarters?

packages/assistant/src/RPA/Assistant/date_picker.py
```diff
@@ -0,0 +1,87 @@
...
+        element = []
+        current_row = []
+        number_of_days = monthrange(self.selected_year, self.selected_month)[1]
+        for day in range(number_of_days):
+            if day % 7 == 0 and day != 0:
+                # element.append(PopupMenuItem(content=Row(current_row, alignment=MainAxisAlignment.SPACE_BETWEEN)))
+                # self.body_element[0].content = Row(current_row, alignment=MainAxisAlignment.SPACE_BETWEEN)
+                current_row = []
+            else:
+                current_row.append(Container(content=Text(day+1), on_click=self._select_day))
```
cmin764, line 50: - Since code might not be self explanatory, a quick comment on the intention here would suffice. Like: `# Displaying the days in the month week by week.`
- The `for` block seems to have no effect since there's commented code. Could this be refactored or at least stripped away? (otherwise a TODO/FIXME + Issue might suffice)

packages/assistant/src/RPA/Assistant/date_picker.py
```diff
@@ -0,0 +1,87 @@
...
+        # self.body_element = element
+        self.update()
+
+    def __init__(self):
+        # initialise default values
+        self.selected_year = datetime.now().year
+        self.selected_month = datetime.now().month
+        self.selected_day = datetime.now().day
+        self.selected_date = None
+        months = (("Jan", "Feb", "Mar", "Apr"), ("May", "Jun", "Jul", "Aug"), ("Sep", "Oct", "Nov", "Dec"))
```
cmin764, line 61: Shall we have this as a class constant since it looks like a duplicate from `_select_month` method?

packages/assistant/src/RPA/Assistant/flet_client.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def _make_flet_event_handler(self, name: str):
+        def change_listener(e: FletEvent):
+            self.results[name] = e.data
+            e.page.update()
+
+        return change_listener
+
+    def add_element(self, element: flet.Control, name: Optional[str] = None):
+        # TODO: validate that element "name" is unique
```
cmin764, line 119: All of these TODOs can be collected as bullet points in a follow-up improvement _epic_ GitHub Issue.

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -0,0 +1,988 @@
...
+        options, default = to_options(options, default)
+
+        options: List[Control] = list(map(Option, options))
+
+        dropdown = Dropdown(options=options, value=default)
+
+        self._client.add_element(Text(value=label))
+        self._client.add_element(dropdown, name=str(name))
+
+    # FIXME: Add keyword back when fixed
```
cmin764, line 691: Would help to provide more info on what is not working so it has to be fixed before uncommenting this.

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -0,0 +1,988 @@
...
+
+
+        """
+
+        def on_click(event: ControlEvent):
+            if isinstance(function, Callable):
+                try:
+                    function(self._client.results)
+                except Exception as err:
+                    print(f"on_click error with button labeled {label}")
```
cmin764, line 979: Wouldn't a logger log message be more appropriate here instead of `print`s?

packages/assistant/src/RPA/Assistant/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    if default is None:
+        default = opts[0]
+
+    if default not in opts:
+        raise ValueError(f"Default '{default}' is not in available options")
+
+    return opts, default
+
+
+def optional_str(val: Any) -> Union[str, None]:
```
cmin764, line 34: Nit:
```suggestion
def optional_str(val: Any) -> Optional[str]:
```
has the same effect. (so with the others below)

cmin764 (approved):
Thank you for this huge contribution!

As the PR is very long, I'm far from having a thorough review here so I pointed out mostly nitpicks and recommendations. (no critical bugs observed so far)

Beside the comments, a few other general points to look over:
1. This package can be aliged a little bit more with the others by following the same structure for tests: `tests/python`, `tests/robot`, `tests/resources`.
   a. Might be even more impactful to have some robot tests as well to run in the CI in RF style.
2. A new package should be published in PyPI (if not already), where the [robocorp](https://pypi.org/user/robocorp/) account and core maintainers should be added to be able to maintain the project.
3. A new entry like _.github/workflows/assistant.yaml_ has to be created for CI actions.
4. Check if docs can be built for this package or if we miss something for that. (`inv docs.build-docs` & `inv docs.host-local-docs` from _meta_)
5. Release notes given this new component once it is released and included in **main** by default (_pyproject.toml_ dep) if that's the plan.

packages/assistant/src/RPA/Assistant/date_picker.py
bogdancondurache: Good point, I'll have to run the related invoke commands for docs and linting

packages/assistant/src/RPA/Assistant/date_picker.py
bogdancondurache: this is also for displaying purposes, they will be displayed 4 on a row in 3 rows

packages/assistant/src/RPA/Assistant/date_picker.py
bogdancondurache: yes, my bad, the reason datepicker is not included (the keyword is commented out) is that it still has some bugs, I forgot to add a FIXME here as well

packages/assistant/src/RPA/Assistant/date_picker.py
bogdancondurache: think this is not even needed and when moving it I missed deleting the old one, will do some cleanup 👍

packages/assistant/src/RPA/Assistant/flet_client.py
bogdancondurache: great idea, will do!

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -0,0 +1,991 @@
...
+            # Every file input result is a list of paths
+            ${result}=    Run dialog
+            FOR    ${path}    IN    @{result.multiple}
+                Log    Selected file: ${path}
+            END
+        """
+
+        def on_pick_result(e: FilePickerResultEvent):
+            if e.files:
+                self._client.results[str(name)] = list(map(lambda f: f.path, e.files))
```
osrjv, line 623: map & filter are discouraged these days, use list comprehensions
```suggestion
                self._client.results[str(name)] = [f.path for f in e.files]
```

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -0,0 +1,991 @@
...
+        .. code-block:: robotframework
+
+            Add heading    An error occurred
+            Add text       See link for documentation
+            Add link       https://robocorp.com/docs    label=Troubleshooting
+            Run dialog
+        """
+        if not label:
+            label = url
+        self._client.add_element(Markdown(f"[{label}]({url})"))
```
osrjv, line 273: Would this benefit from escaping the content?

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -0,0 +1,991 @@
...
+        """
+        if not isinstance(size, Size):
+            size = Size(size)
+
+        if size == Size.Small:
+            self._client.add_element(element=Text(heading, style="headlineSmall"))
+        elif size == Size.Medium:
+            self._client.add_element(element=Text(heading, style="headlineMedium"))
+        elif size == Size.Large:
+            self._client.add_element(element=Text(heading, style="headlineLarge"))
```
osrjv, line 212: This (and other similar methods) would maybe benefit from only mapping the style string in these conditionals, and having a single add_element call, just to make reading and refactoring easier. It can also be done with a single dict lookup, if so inclined:
```python
style = {
    Size.Small: "headlineSmall",
    Size.Medium: "headlineMedium",
    Size.Large: "headlineLarge",
}[size]

self._client.add_element(element=Text(heading, style=style))
```

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -0,0 +1,991 @@
...
+from robot.libraries.BuiltIn import BuiltIn, RobotNotRunningError
+from RPA.Assistant.flet_client import FletClient
+
+from .dialog_types import Icon, Options, Result, Size
+from .utils import optional_str, to_options
+
+
+@library(scope="GLOBAL", doc_format="REST", auto_keywords=False)
+class Assistant:
+    """This library is still in Beta and needs more testing and features.
```
osrjv, line 40: I wouldn't start the library documentation with this 😅 It can be a note later, or a highlighted banner in rpaframework.org.

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -0,0 +1,991 @@
...
+            Add text input    message
+            ...    label=Feedback
+            ...    placeholder=Enter feedback here
+            ...    rows=5
+            ${result}=    Run dialog
+            Send feedback message    ${result.email}  ${result.message}
+        """
+        # TODO: Do this in a cleaner way. Workaround because we use on_change
+        # handlers to record values, so default value otherwise will be missed
+        self._client.results[name] = placeholder
```
osrjv, line 487: Should there be some guarding (even if just warning logs) that multiple fields don't set the same result name?

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -0,0 +1,991 @@
...
+            Add file input    name=types     file_type=PDF files (*.pdf)
+
+            # Every file input result is a list of paths
+            ${result}=    Run dialog
+            FOR    ${path}    IN    @{result.multiple}
+                Log    Selected file: ${path}
+            END
+        """
+
+        def on_pick_result(e: FilePickerResultEvent):
```
osrjv, line 621: I think (or hope) pylint/flake8 will complain about these single char variable names

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -0,0 +1,991 @@
...
+            Add radio buttons
+            ...    name=user_type
+            ...    options=Admin,Maintainer,Operator
+            ...    default=Operator
+            ...    label=User type
+            ${result}=      Run dialog
+            Log    User type should be: ${result.user_type}
+        """
+        options, default = to_options(options, default)
+        radios: List[Control] = list(
```
osrjv, line 762: List comprehensions here as well, imo

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -0,0 +1,991 @@
...
+            Log    The username is: ${result.username}
+        """
+
+        # FIXME: support timeout
+
+        self._client.display_flet_window(title, height, width, on_top)
+        return self._client.results
+
+    @keyword("Ask User", tags=["dialog"])
+    def ask_user(self, timeout: int = 180, **options: Any) -> Result:
```
osrjv, line 886: This keyword and what it does confuses me a bit. Or rather, I understand from looking at the code, but reading the documentation I'd be a bit baffled. Maybe rethink if this is required, and if it's a very useful shorthand think about integrating it somehow into the normal dialog configuration/running.

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -0,0 +1,991 @@
...
+        """
+
+        # FIXME: support timeout
+
+        self.add_submit_buttons(["Submit", "Close"], "Submit")
+        self._client.display_flet_window(**options)
+        return self._client.results
+
+    @keyword("Refresh", tags=["dialog"])
+    def refresh(self):
```
osrjv, line 916: Rename this to something less generic, would confuse a lot in a larget project. Maybe refresh dialog?

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -0,0 +1,991 @@
...
+        """Can be used to update UI elements when adding elements while dialog is running"""
+        if self._client.page:
+            self._client.update_elements(self._client.page)
+        else:
+            raise Exception("No dialog open")
```
osrjv, line 921: Don't raise base Exceptions, turn it into a RuntimeError or something

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -0,0 +1,991 @@
...
+                try:
+                    BuiltIn().run_keyword(function, *args, **kwargs)
+                except Exception as err:
+                    print(f"on_click error with button labeled {label}")
+                    print(err)
+
+        self._client.add_element(ElevatedButton(label, on_click=on_click))
+
+    @keyword("Add Next Ui Button", tags=["dialog"])
+    def add_next_ui_button(self, label: str, function: Union[Callable, str]):
```
osrjv, line 953: Rethink the naming here a bit. The term "next UI" doesn't mean anything to me. Maybe we should call it switching views, or something to that effect

packages/assistant/src/RPA/Assistant/utils.py
bogdancondurache: done

packages/assistant/src/RPA/Assistant/library.py
bogdancondurache: done, inside the class

packages/assistant/src/RPA/Assistant/date_picker.py
bogdancondurache: done

packages/assistant/src/RPA/Assistant/date_picker.py
bogdancondurache: done

packages/assistant/src/RPA/Assistant/library.py
bogdancondurache: not sure yet, I'll check this and also what library we use already for escaping (if any) as Flet doesn't include it

packages/assistant/src/RPA/Assistant/library.py
bogdancondurache: done

packages/assistant/src/RPA/Assistant/library.py
bogdancondurache: yep, totally agreed with this, wanted to refactor it like this but kind of postponed all refactoring until after the release

bogdancondurache: done

packages/assistant/src/RPA/Assistant/library.py
bogdancondurache: not sure how to have highlighted content on rpaframework.org and also we discussed that the Beta note should be at the beginning for users to know exactly why they shouldn't use it without our guidance. This is based on other discussions had with @kariharju and @xylix so I wouldn't change this right now

packages/assistant/src/RPA/Assistant/library.py
bogdancondurache: yes, we will add the guarding later on

packages/assistant/src/RPA/Assistant/library.py
bogdancondurache: done

packages/assistant/src/RPA/Assistant/library.py
bogdancondurache: agreed, done

packages/assistant/src/RPA/Assistant/library.py
bogdancondurache: It should be the default way of using Assistant dialogs (instead of run dialog) for simple cases, like forms. Agreed that the documentation could be better, but don't think that's the biggest priority right now, so maybe we can discuss about the docs later on. And yes, this keyword was a requirement

packages/assistant/src/RPA/Assistant/library.py
bogdancondurache: done

packages/assistant/src/RPA/Assistant/library.py
bogdancondurache: agreed, done

packages/assistant/src/RPA/Assistant/library.py
bogdancondurache: yes, this was already discussed needs a better naming, but we're not sure about which one yet 😄

packages/assistant/src/RPA/Assistant/flet_client.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def resolve_absolute_position(
+    location: Union[Location, Tuple],
+) -> Tuple[int, int]:
+
+    if location is Location.TopLeft:
+        left_coordinate = 0
+        top_coordinate = 0
+        return (left_coordinate, top_coordinate)
```
osrjv, line 21: tbh I'd almost say this is more readable
```python
if location is Location.TopLeft:
    return (0, 0)
```

packages/assistant/src/RPA/Assistant/flet_client.py
```diff
@@ -0,0 +1,183 @@
...
+    elif isinstance(location, tuple):
+        return location
+    else:
+        raise ValueError("Locations other than Center or TopLeft are not yet supported")
```
osrjv, line 25: This error message doesn't match the content

packages/assistant/src/RPA/Assistant/flet_client.py
```diff
@@ -0,0 +1,183 @@
...
+    control: Control
+    page: Page
+
+
+class FletClient:
+    """Class for wrapping flet operations"""
+
+    def __init__(self) -> None:
+        self._conn = self._preload_flet()
+        self.elements: List[List[Control]] = [[]]
```
osrjv, line 41: On a quick skim: Why the 2d array? What is a list of controls? An element?

packages/assistant/src/RPA/Assistant/flet_client.py
```diff
@@ -0,0 +1,183 @@
...
+        self.invisible_elements: List[List[Control]] = [[]]
+        self.results: Result = {}
+        self._pagination = 0
+        self.page: Optional[Page] = None
+        self._fvp = None
```
osrjv, line 46: Are the non-underscore props actually public or what is the logic here, seems pretty random

osrjv: sorry, attributes, wrong language

packages/assistant/src/RPA/Assistant/flet_client.py
```diff
@@ -0,0 +1,183 @@
...
+        return
+
+    def update_elements(self, page: Page):
+        """Updates the UI and shows new elements which have been added into the element lists"""
+        return self._execute(page)()
+
+    def flet_update(self):
+        """Runs a plain update of the flet UI, updating existing elements"""
+        if not self.page:
+            raise Exception("Flet update called when page is not open")
```
osrjv, line 182: Please don't raise plain Exceptions, maybe a RuntimeError at least?

packages/assistant/tests/robot/assistant_test.robot
```diff
@@ -0,0 +1,62 @@
...
+        Log to Console  ${result}
+        Log to Console  ${result.stdout}
+
+
+Dummy Process
+    Set Log Level  TRACE
+    ${result}=  Run Process  echo  0
+    Log to Console  ${result}
+    Log to Console  ${result.stdout}
+
```
osrjv, line 31: 👀 Please use robotframework-tidy

packages/assistant/src/RPA/Assistant/flet_client.py
```diff
@@ -0,0 +1,189 @@
...
+    def current_invisible_elements(self):
+        return self._invisible_elements[self._pagination]
+
+    def _cleanup(self) -> None:
+        # Source: https://github.com/flet-dev/flet/blob/89364edec81f0f9591a37bdba5f704215badb0d3/sdk/python/flet/flet.py#L146
+        self._conn.close()
+        if self._fvp is not None and not is_windows():
+            try:
+                logging.debug(f"Flet View process {self._fvp.pid}")
+                os.kill(self._fvp.pid + 1, signal.SIGKILL)
```
xylix, line 65: Feedback from Jippo and Fabio: This is actually pretty dangerous and may kill a random process. If we want to kill a process and all it's children LSP had some code we could take inspiration from here: https://github.com/robocorp/robotframework-lsp/blob/f485ec493109311f32df08a7e0b2e243ed974d5c/robocorp-python-ls-core/src/robocorp_ls_core/basic.py#L220

packages/assistant/src/RPA/Assistant/flet_client.py
xylix: resolved in 08b33446703e680bb0ceae9c53629e23cd3eb65b

packages/assistant/src/RPA/Assistant/flet_client.py
xylix: fixed in 5cfb75d6, this was not anymore used so the 2d-ness has now been removed.

PR #685, opened 2022-11-03
`RPA.HTTP` - Add `Check Vulnerabilities` keyword

mikahanninen:
The keyword will check if the installed OpenSSL version is within the range of 3.0.0 - 3.0.6 and outputs a warning message in this case.

Outputs all results as a list of dictionaries containing type, vulnerable status (True if there are issues) and the message (if in the future additional checks would be made those will be included in the same list).

usage example in Robot Framework
```robotframework
${results}=    Check Vulnerabilities
FOR    ${result}    IN    @{results}
    Log To Console    TYPE: ${result}[type]
    Log To Console    VULNERABLE: ${result}[vulnerable]
    Log To Console    MESSAGE: ${result}[message]
END
```

example output from the above code, with GOOD OpenSSL version
```bash
TYPE: OpenSSL
VULNERABLE: False
MESSAGE: OpenSSL 1.1.1n  15 Mar 2022
```

example warning message and the output from the above code, with **VULNERABLE** OpenSSL version
`[ WARN ] Dependency with HIGH severity vulnerability detected: 'OpenSSL 3.0.5 5 Jul 2022'. For more information see https://robocorp.com/docs/faq/openssl-cve-2022-11-01`
```bash
TYPE: OpenSSL
VULNERABLE: True
MESSAGE: Dependency with HIGH severity vulnerability detected: 'OpenSSL 3.0.5 5 Jul 2022'. For more information see https://robocorp.com/docs/faq/openssl-cve-2022-11-01
```

### DIFF
--- a/packages/main/src/RPA/HTTP.py
+++ b/packages/main/src/RPA/HTTP.py
@@ -1,6 +1,7 @@
 # pylint: disable=C0411,C0412,C0413
 import logging
 from pathlib import Path
+import re
 from typing import Any, Optional, Union
 from urllib.parse import urlparse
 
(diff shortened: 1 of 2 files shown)


packages/main/src/RPA/HTTP.py
```diff
@@ -258,3 +259,39 @@ def download(
         self._create_or_overwrite_target_file(dirname / filename, response, overwrite)
 
         return response
+
+    def check_vulnerabilities(self):
+        """Check for possible Python vulnerabilities in the installed
```
bogdancondurache, line 264:
```suggestion
        """Check for possible vulnerabilities in the installed
```
I don't think they are only Python related

mikahanninen: Good point. Changed.

packages/main/src/RPA/HTTP.py
```diff
@@ -258,3 +259,52 @@ def download(
...
+            {"type": "OpenSSL", "vulnerable": vulnerable, "message": message}
+        )
+        if vulnerable:
+            self.logger.warning(message)
+        return all_results
+
+    def _check_openssl_vulnerabilities(self):
+        message = "No OpenSSL detected"
+        try:
+            import ssl  # pylint: disable=C0415
```
xylix, line 295: I think this means that we only check openssl version if pyOpenSSL is also present. I think it makes sense, but I wonder if there is some way to document it? But if this does not seem important to other reviewers I think it's fine to leave as is.

PR #687, opened 2022-11-04
Reorder imports so `win32api` is imported before `pywintypes`

blakewaud:
This resolves an import error related to `pywintypes39.dll`


packages/main/src/RPA/Outlook/Application.py
```diff
@@ -11,8 +11,8 @@
 from RPA.Email.common import counter_duplicate_path
 
 if platform.system() == "Windows":
-    import pywintypes
     import win32api
+    import pywintypes
     import win32com.client
```
mikahanninen, line 16: maybe more this also before pywintypes, like in the Excel.Application

mikahanninen (approved):

Issue #626, opened 2022-09-13
`RPA.Windows` keyword `Get elements` not retrieving all elements


blakewaud:
Keyword failed to retrieve all elements based on common locator. If you review the print tree, there are multiple RadioButtonControl types with the name "No" but the `Get elements` only returns one item.

```
KEYWORD win . Print Tree ${ANCHOR}
Documentation:	
Print Control element tree.

Start / End / Elapsed:	20220913 16:10:10.173 / 20220913 16:10:10.679 / 00:00:00.506
16:10:10.175	TRACE	Arguments: [ 'id:"individual-leftcontent"' ]	
16:10:10.177	INFO	Getting element with locator: id:"individual-leftcontent"	
16:10:10.242	DEBUG	argument root = None	
16:10:10.244	DEBUG	active anchor = None	
16:10:10.253	DEBUG	active window = ControlType: PaneControl    ClassName: Chrome_WidgetWin_1    AutomationId:     Rect: (708,328,2888,1698)[2180x1370]    Name: 'EIN Individual Request - Online Application - Google Chrome'    Handle: 0x7005FA(7341562)	
16:10:10.259	DEBUG	resulting root = ControlType: PaneControl    ClassName: Chrome_WidgetWin_1    AutomationId:     Rect: (708,328,2888,1698)[2180x1370]    Name: 'EIN Individual Request - Online Application - Google Chrome'    Handle: 0x7005FA(7341562)	
16:10:10.259	INFO	Root element: <uiautomation.uiautomation.PaneControl object at 0x0000021B64402760>	
16:10:10.260	INFO	Locator 'id:"individual-leftcontent"' produced matcher: MatchObject(locators=[('AutomationId', 'individual-leftcontent', 0)], _classes=set(), max_level=0)	
16:10:10.422	INFO	Returning element: WindowsElement(item=<uiautomation.uiautomation.GroupControl object at 0x0000021B55FEB610>, locator='id:"individual-leftcontent"', name='', automation_id='individual-leftcontent', control_type='GroupControl', class_name='', left=1235, right=2079, top=707, bottom=1345, width=844, height=638, xcenter=1657, ycenter=1026)	
16:10:10.426	INFO	ControlType: GroupControl    ClassName:     AutomationId: individual-leftcontent    Rect: (1235,707,2079,1345)[844x638]    Name: ''    Handle: 0x0(0)	
[...]

Issue #371, opened 2022-01-31
RPA.Windows 2.0.0 cannot find menu_items past a certain depth


orlof:
![image](https://user-images.githubusercontent.com/1869202/151804379-514e66e2-267a-403b-89e8-d4a419707f80.png)

mikahanninen:
Possible bug, but needs to be investigated further.

cmin764:
For such "Zoom in" menu item to be visible in the tree, we should expand the tree first.

`Get Element     Zoom in` alone can't retrieve that item, but if we surf until it, we can:
```robotframework
Retrieve Nested Notepad Elements
    [Documentation]     Try to unfold menu items until being able to retrieve a deep
    ...     element inside the tree as soon as it becomes visible.
    [Setup]   Windows Run    Notepad

    Control Window      subname:Notepad
    Click   View
    Click   Zoom
    ${zoom_in} =   Get Element     Zoom in
    Log To Console      "Zoom in" item: ${zoom_in}

    [Teardown]  Close Current Window
```

**Zoom in** item
```
WindowsElement(item=<uiautomation.uiautomation.MenuItemControl object at 0x00000210BBA01070>, locator='Zoom in', name='Zoom in', automation_id='', control_type='MenuItemControl', class_name='MenuFlyoutItem', left=1581, right=2025, top=384, bottom=442, width=444, height=58, xcenter=1803, ycenter=413)
```

PR #693, opened 2022-11-14
Windows: Get all the elements & fix depth

cmin764:
- [ ] `Get Elements   all=${True}`: https://github.com/robocorp/rpaframework/issues/626
- [ ] Depth: https://github.com/robocorp/rpaframework/issues/371

### DIFF
--- a/packages/core/poetry.lock
+++ b/packages/core/poetry.lock
@@ -137,7 +137,7 @@ toml = ["tomli"]
 
 [[package]]
 name = "cryptography"
-version = "38.0.1"
+version = "38.0.3"
 description = "cryptography is a package which provides cryptographic recipes and primitives to Python developers."
 category = "dev"
 optional = false
@@ -164,7 +164,7 @@ python-versions = ">=3.7"
 
 [[package]]
 name = "exceptiongroup"
-version = "1.0.0rc9"
+version = "1.0.2"
 description = "Backport of PEP 654 (exception groups)"
 category = "main"
 optional = false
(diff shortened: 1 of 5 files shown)


cmin764:
Use cases of `Get Elements` on **Calculator** (Windows 11) given the new `siblings_only` introduced param:

1. `class:Button regex:.*o.*` -> finds first button which contains an `o` in name, under the current active window, then returns all the buttons on the same level with the found one if and only if the others contain an `o` as well
Screenshot 2022-11-16 at 17 09 22

2. `class:Button regex:.*o.*   siblings_only=${False}` -> the same as **1.**, but return this time ALL the buttons found in the entire active window which contains an `o` under the name
Screenshot 2022-11-16 at 17 11 47

### Remarks
- `root_element` param stands as root for both the initial element search and for the tree root search when filtering out all candidate elements to be returned.
   - If this is not provided, then the current set anchor or window (in this order) is assumed. (and Desktop root if none of previous were set -- the exact same logic as with `Get Element`)
- If `locator` contains multiple levels for a narrower search (`id:NumberPad > class:Button`) only the last level (`class:Button`) is taken into account for establishing the filtering criteria of matching candidates.
   - So a `id:NumberPad > class:Button regex:.*o.*` locator will take all the NUMERIC buttons containing an `o` in their name (Two, Zero etc.).
   - If this gets used with `siblings_only=${False}` then all the buttons having an `o` (so not just the numeric ones) will be returned.
   - To change the behaviour above (like norrowing further the search, but still taking into account all possible elements), just provide a `root_element`.

packages/windows/src/RPA/Windows/keywords/elements.py
```diff
@@ -80,7 +80,6 @@ def print_tree(
 
         :param locator: string locator or Control element
         :param max_depth: maximum depth level (defaults to 8)
-        :param encoding: defaults to "utf-8"
```
cmin764, line 83: This referenced a param that was no longer existing.

PR #709, opened 2022-11-24
Updated version numbers

TimothyLopas:


packages/windows/src/RPA/Windows/__init__.py
```diff
@@ -41,10 +41,10 @@ class Windows(WindowsElementsMixin, DynamicCore):
         channels:
```
cmin764, line 41: I would refactor from above this as well:

From
```
Ultimately this library will replace current library ``RPA.Desktop.Windows`` if seen as approriate.
```

To (or similar)
```
This library replaces the old deprecated ``RPA.Desktop.Windows``, so please use ``RPA.Windows`` instead.
```

packages/windows/src/RPA/Windows/__init__.py
```diff
@@ -41,10 +41,10 @@ class Windows(WindowsElementsMixin, DynamicCore):
...
           - conda-forge
         dependencies:
-          - python=3.7.5
-          - pip=20.1
+          - python=3.9.13
+          - pip=22.1.2
           - pip:
-            - rpaframework-windows==2.0.0
+            - rpaframework-windows==6.2.0
```
cmin764, line 47:
```suggestion
            - rpaframework-windows==6.2.0  # just the Windows library (`rpaframework` alone includes it)
```

cmin764 (approved):
LGTM!

Left a few related improvements, feel free to merge afterwards without a release note as this is just documentation. (that automatically builds on _master_)

packages/windows/src/RPA/Windows/__init__.py
TimothyLopas: Change made

packages/windows/src/RPA/Windows/__init__.py
TimothyLopas: change made

Issue #720, opened 2022-12-02
`RPA.Browser.*` library improvements (and Portal examples)


cmin764:
## Frequent issues coming from the support side
- [ ] https://github.com/robocorp/rpaframework/issues/341 ([comment](https://github.com/robocorp/rpaframework/issues/341#issuecomment-1203815213))
- [ ] Portal example on selecting a certificate during the browser window prompt.
- [ ] Portal example with switching between windows & tabs. (@TejusVenkateshRobocorp )
- [ ] ~Add an example showcasing the `proxy` usage with the **OAB** keyword. (write a robot test for it as well to check behavior)~ (`proxy` is simply passed during library import)
- [x] Playwright external links: https://github.com/robocorp/rpaframework/issues/747
- [x] IE support: https://github.com/robocorp/rpaframework/issues/745
- [ ] Use Selenium's web-driver manager: https://github.com/robocorp/rpaframework/issues/777
- [x] #781
- [ ] Enable custom paths for the web-driver binary download location in _core_ (@mikahanninen )
- [x] Custom download: https://github.com/robocorp/rpaframework/issues/412

#### Selenium
- [x] https://github.com/robocorp/rpaframework/issues/883 🐛 
- [x] https://github.com/robocorp/rpaframework/issues/586 🐛 
- [x] #919
- [x] #920

#### Selenium
- [x] https://github.com/robocorp/rpaframework/issues/882
- [x] https://github.com/robocorp/rpaframework/issues/865
- [x] https://github.com/robocorp/rpaframework/issues/884
- [x] https://github.com/robocorp/rpaframework/issues/762
- [x] #940
- [x] #939
- [x] Fix `headless` unit-testing

#### Selenium
- [x] https://github.com/robocorp/rpaframework/issues/949 🐛 
- [x] #941

#### Playwright
- [x] #942
- [x] https://github.com/robocorp/rpaframework/issues/166
- [x] https://github.com/robocorp/rpaframework/issues/188

TejusVenkateshRobocorp:
Can we look at the Open Available Browser to use alias parameter to help in the Switch Windows keyword? :)

cmin764:
> Can we look at the Open Available Browser to use alias parameter to help in the Switch Windows keyword? :)

Definitely we can, the question is when. We can sync with Tommi on priorities

cmin764:
I conclude this epic done now, will start another one with the rest of the unticked checkboxes when we'll consider those a priority again.

PR #780, opened 2023-01-23
Use RPA_SELENIUM_BROWSER_ORDER env variable to determine default browsers

avaissi:
* ~~ROBOCORP_BROWSERS~~ RPA_SELENIUM_BROWSER_ORDER env variable, if available, is used to determine default browsers
* ~~ROBOCORP_BROWSERS~~ RPA_SELENIUM_BROWSER_ORDER env is re-checked each time Open Available Browser is called

### DIFF
--- a/docs/source/releasenotes.rst
+++ b/docs/source/releasenotes.rst
@@ -9,6 +9,12 @@ Release notes
 `Released <https://pypi.org/project/rpaframework/#history>`_
 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
+20.1.1 - 25 Jan 2023
+--------------------
+
+- Library **RPA.Browser.Selenium** (:issue:`781`): Ability to determine the
+  default search order of browsers, when using the ``Open Available Browser``-keyword
+
 20.1.0 - 19 Jan 2023
 --------------------
 
--- a/docs/source/releasenotes.rst
+++ b/docs/source/releasenotes.rst
@@ -13,7 +13,7 @@ Release notes
 --------------------
 
 - Library **RPA.Browser.Selenium** (:issue:`781`): Ability to determine the
-  default search order of browsers, when using the ``Open Available Browser``-keyword
+  default search order of browsers, when using the ``Open Available Browser`` keyword.
 
 20.1.0 - 19 Jan 2023
 --------------------
--- a/packages/main/src/RPA/Browser/Selenium.py
+++ b/packages/main/src/RPA/Browser/Selenium.py
@@ -73,7 +73,7 @@ def browser_preferences() -> Dict[str, List[str]]:
         preferences = {
             "default": [
                 browser.strip()
-                for browser in os.getenv("RPA_SELENIUM_BROWSER_ORDER", "").split(sep=",")
+                for browser in browsers.split(sep=",")
             ],
         }
     else:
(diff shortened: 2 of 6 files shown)


avaissi:
The reasoning behind the PR is give control for user to determine the order default browsers are tried

cmin764 (commented):
This looks good in general! An additional improvement would be to move the browser preference out of the _core_ package entirely.

Is this old `DRIVER_PREFERENCE` used in any other place apart from `RPA.Browser.Selenium`? If not, then maybe it would be best to move this logic near the library itself in _main_. What do you think?

avaissi:
> This looks good in general! An additional improvement would be to move the browser preference function out of the _core_ package entirely, into _main_.
> 
> Is this old `DRIVER_PREFERENCE` constant used in any other place apart from `RPA.Browser.Selenium`? If not, then maybe it would be best to move this logic near the library itself in _main_ (so we don't depend on _core_ updates for such behaviour). What do you think?
> 
> Newly created [Issue](https://github.com/robocorp/rpaframework/issues/781) from [epic](https://github.com/robocorp/rpaframework/issues/720)

This sounds good to me. Less dependencies the better 👍 I'll update the PR along with possible variable name change

cmin764 (changes requested):
This looks better now!

Further required changes (which I guess they are already WIP since this PR is still marked as draft):
1. Bump the major.minor.**hotfix** version of _core_, even if it looks like a breaking change, since only _main_ depends on this change and is already made independent of it in this PR.
2. Bump the hotfix version as well of _main_.
3. Add release note into for main into _docs/source/releasenotes.rst_.
4. You can merge.

### Post-merge
1. Release _core_ from master.
2. Poetry update _main_ (to use latest just released _core_) and release _main_.
3. Release published with releasenotes.io and Slack announcement channels. (with release notes taken from rpaframework.org)

docs/source/releasenotes.rst
```diff
@@ -9,6 +9,12 @@ Release notes
 `Released <https://pypi.org/project/rpaframework/#history>`_
 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
+20.1.1 - 25 Jan 2023
+--------------------
+
+- Library **RPA.Browser.Selenium** (:issue:`781`): Ability to determine the
+  default search order of browsers, when using the ``Open Available Browser``-keyword
```
cmin764, line 16:
```suggestion
  default search order of browsers, when using the ``Open Available Browser`` keyword.
```

maybe?

packages/main/src/RPA/Browser/Selenium.py
```diff
@@ -64,6 +64,28 @@ def ensure_scheme(url: str, default: Optional[str]) -> str:
...
+def browser_preferences() -> Dict[str, List[str]]:
+    """Get lists of browser preferences for OS.
+    Prefers environment variable RPA_SELENIUM_BROWSER_ORDER, if defined.
+    """
+    browsers = os.getenv("RPA_SELENIUM_BROWSER_ORDER", "")
+    if browsers:
+        preferences = {
+            "default": [
+                browser.strip()
+                for browser in os.getenv("RPA_SELENIUM_BROWSER_ORDER", "").split(sep=",")
```
cmin764, line 76:
```suggestion
                for browser in browsers.split(sep=",")
```
DRY

cmin764 (approved):
LGTM! (left some nits)

PR #782, opened 2023-01-25
Add openai library

orlof:


orlof:
Decided to put in a separate package.

PR #786, opened 2023-01-26
Fix semantic versioning break

osrjv:
Core `10.1.1` removed a property that was referenced in `RPA.Browser.Selenium`, however old versions of `rpaframework` only pinned the major version for core. The previous version was yanked, and this will restore compatiblity if released.

Not tested, since I am unable to run the new invocations on my machine, and too busy to debug 😇 
Please make sure I didn't typo anything in the revert.


avaissi (approved):
Tested with old rpaframework main (version 17.7.0) with fixed core. Looks good to me 👍

PR #792, opened 2023-01-31
New library OpenAI

orlof:
separate package due to python incompatibility with rpafw

### DIFF
--- a/docs/source/libraries/openai/index.rst
+++ b/docs/source/libraries/openai/index.rst
@@ -1,0 +1,24 @@
+######
+OpenAI
+######
+
+A library that wraps the `openai package`_, giving robots the ability to access the Open AI Microsoft API programmatically.
+
+.. _openai package: https://pypi.org/project/openai
+
+.. toctree::
+   :maxdepth: 1
+   :hidden:
+
+   python
+
+Keywords
+========
+
+🔗 Direct link to `keyword documentation <../../libdoc/RPA_OpenAI.html>`_.
+
+-------
+
+.. raw:: html
+
+   <iframe scrolling="no" id="libdoc" src="../../libdoc/RPA_OpenAI.html"></iframe>
--- a/docs/source/libraries/openai/python.rst
+++ b/docs/source/libraries/openai/python.rst
@@ -1,0 +1,13 @@
+##########
+Python API
+##########
+
+******
+OpenAI
+******
+
+.. automodule:: RPA.OpenAI
+   :members:
+   :inherited-members:
+   :undoc-members:
+   :show-inheritance:
--- a/packages/openai/src/RPA/OpenAI.py
+++ b/packages/openai/src/RPA/OpenAI.py
@@ -1,0 +1,256 @@
+import logging
+from typing import Optional
+from robot.api.deco import keyword
+import openai
+
+
+class OpenAI:
+    """Library to support `OpenAI <https://openai.com>`_ service.
+
+    **Robot Framework example usage**
+
+    .. code-block:: robotframework
+
+        *** Settings ***
+        Library    RPA.Robocorp.Vault
+        Library    RPA.OpenAI
+
+        *** Tasks ***
+        Create a text completion
+            ${secrets}   Get Secret   secret_name=OpenAI
+            Authorize To OpenAI   api_key=${secrets}[key]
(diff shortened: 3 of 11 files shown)


packages/openai/src/RPA/OpenAI.py
```diff
@@ -0,0 +1,241 @@
...
+        temperature: Optional[int] = 0.7,
+        max_tokens: Optional[int] = 256,  
+        top_p: Optional[int] = 1,   
+        frequency_penalty: Optional[int] = 0,   
+        presence_penalty: Optional[int] = 0,
+        result_format: Optional[str] = "string"      
+    ) -> None:
+        """Keyword for creating text completions in OpenAI. Keyword returns a text string.
+
+        :param prompt: Text submitted to OpenAI for creating natural language.
```
mikahanninen, line 87: All parameters are not described.

packages/openai/src/RPA/OpenAI.py
```diff
@@ -0,0 +1,241 @@
...
+        """
+        openai.api_key = api_key
+
+    @keyword
+    def completion_create(self, 
+        prompt: str,
+        model: Optional[str] = "text-davinci-003",
+        temperature: Optional[int] = 0.7,
+        max_tokens: Optional[int] = 256,  
+        top_p: Optional[int] = 1,
```
mikahanninen, line 80: Pretty bad parameter name.

The underlying OpenAI can use whatever bad naming it can, but we should not have our keyword users suffer the same way.

packages/openai/src/RPA/OpenAI.py
```diff
@@ -0,0 +1,241 @@
...
+        if result_format == "json":
+            return response
+        else:
+            return None
+
+    @keyword
+    def image_create(self, 
+        prompt: str, 
+        size: Optional[str] = "512x512", 
+        n: Optional[int] = 1,
```
mikahanninen, line 133: bad parameter name, maybe `count` or `number` would be better

packages/openai/src/RPA/OpenAI.py
```diff
@@ -0,0 +1,241 @@
...
+        if result_format == "json":
+            return response
+        else:
+            return None
+
+    @keyword
+    def image_create_variation(self,
+        src_image: str,
+        size: Optional[str] = "512x512", 
+        n: Optional[int] = 1,
```
mikahanninen, line 188: commented on this in previous keywords

packages/openai/src/RPA/OpenAI.py
```diff
@@ -0,0 +1,241 @@
...
+        result_format: Optional[str] = "list"
+        ) -> None:
+
+        """Keyword for creating one or more variations of a image. Keyword returns a list of urls for the images created.
+        Source file must be a valid PNG file, less than 4MB, and square.
+        
+        param size: Size of the files to be created. 256x256, 512x512, 1024x1024
+
```
mikahanninen, line 196: there should not be a need to have empty spaces between `:param: `entries

mikahanninen (changes requested):
As a overall comment, the package should have some tests. At least Robot Framework tests.

docs/source/libraries/openai/index.rst
```diff
@@ -0,0 +1,24 @@
+########
```
mikahanninen, line 1: I think these need to be the same length as the text. 

```rst
######
OpenAI
######
```

docs/source/libraries/openai/python.rst
```diff
@@ -0,0 +1,13 @@
+##########
+Python API
+##########
+
+********
```
mikahanninen, line 5: similar comment here about the length of these "decorator" characters, in here `*` need to match length of the text within

packages/openai/src/RPA/OpenAI.py
orlof: ChatGPT told me that num_images is the best argument name for this :)

packages/openai/src/RPA/OpenAI.py
orlof: ...the irony is that `n` comes directly from OpenAI API :-)) fixed

mikahanninen (approved):

PR #793, opened 2023-01-31
Improve PDF docs

cmin764:
Prepares the docs for the main next major release. References Portal [example](https://github.com/robocorp/example-parse-pdf-invoice).


blakewaud (approved):
LGTM 👍

PR #817, opened 2023-02-08
Fix  for docs in RPA.Assistant

bogdancondurache:


mikahanninen (approved):
LGTM

docs/source/releasenotes.rst
```diff
@@ -96,7 +96,6 @@ Release notes
   compatibility mode. (robot example: https://github.com/robocorp/example-ie-mode-edge)
 - Library **RPA.SAP**: Fix error in dependency import order in the underlying
   `SapGuiLibrary`.
->>>>>>> master
```
cmin764, line 99: This already got solved in the latest _master_.

pyproject.toml
```diff
@@ -13,7 +13,7 @@ rpaframework-google = "^7.0.0"
 rpaframework-openai = "^1.0.1"
 rpaframework-recognition = "^5.0.1"
 rpaframework-windows = "^7.0.3"
-rpaframework-assistant = "^1.0.4"
+rpaframework-assistant = "^1.0.5"
```
cmin764, line 16: Better to run a `poetry update -vvv` (or `inv install.update`) directly in _master_ (alongside this change) once you merge and release the new assistant package, so you get it in the docs straight from PyPI.

(as PyPI releases usually work from _master_ post-merge)

cmin764 (approved):
LGTM! (left two nits)

pyproject.toml
xylix: we have been building the pypi releases by hand for rpaframework-assistant, but valid point about the workflow.

bogdancondurache: makes sense, I'll remove this change from the PR and we can do it afterwards then

bogdancondurache:
```suggestion
rpaframework-assistant = "^1.0.4"
```

docs/source/releasenotes.rst
bogdancondurache: shouldn't be an issue with it being here then 😄

PR #822, opened 2023-02-10
RPA.Assistant doesn't have keywords used in example

pasikokko:
`Show dialog` and `Close dialog` keywords are not in the library


xylix (approved):

xylix:
Thanks, good catch! Had missed this. We actually have adding some keyword to support progress-bar use cases on our backlog, but as-is the example doesn't make any sense.

PR #824, opened 2023-02-13
Assistant: Fix default values and docs

bogdancondurache:
What this PR contains:
* Fixed the bugs with the default vales
* Updated the docs to fix some bugs and inaccuracies + migration guide
* Update Assistant to 1.0.6 in root and updates lock files in root and Assistant

### DIFF
--- a/packages/assistant/src/RPA/Assistant/library.py
+++ b/packages/assistant/src/RPA/Assistant/library.py
@@ -51,6 +51,10 @@ class Assistant:
     other robots inside the current one and determine what to display to the
     user based on his previous responses.
 
+    It is not included in the `rpaframework` package, so in order to use it
+    you have to add `rpaframework-assistant` with the desired version in your
+    *conda.yaml* file
+
     Some examples of use-cases could be the following:
 
     - Displaying generated files after an execution is finished
@@ -585,7 +589,6 @@ class Assistant:
         name: str,
         label: Optional[str] = None,
         source: Optional[str] = None,
-        destination: Optional[str] = None,
         file_type: Optional[str] = None,
         multiple: bool = False,
     ) -> None:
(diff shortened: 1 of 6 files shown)


mikahanninen (approved):

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -585,7 +589,6 @@ def add_file_input(
         name: str,
         label: Optional[str] = None,
         source: Optional[str] = None,
-        destination: Optional[str] = None,
```
cmin764, line 588: These kinds of changes are usually breaking changes. Still, since this is under active dev and the `destination` param was never meant to work, I guess is not that bad.

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -1081,7 +1089,7 @@ def add_next_ui_button(self, label: str, function: Union[Callable, str]):
 
         :param label: Text for the button
         :param function: Python function or Robot Keyword name, that will take form
-        results as it's first argument
+            results as it's first argument
```
cmin764, line 1092: Nit: `results as its first argument`

cmin764 (commented):
Late at the party, just left my two cents.

packages/assistant/src/RPA/Assistant/library.py
xylix: Fixes this in future PR at here https://github.com/robocorp/rpaframework/pull/830

packages/assistant/src/RPA/Assistant/library.py
xylix: I think the important part here is that it was not documented to do anything and was not supposed to do anything, so the existence of the parameter was a _bug_ here. Or that's how I see it. I guess one could semver bikeshed a bunch about this but yeah it's "not that bad". But also just bumping the major would not be that bad.

PR #829, opened 2023-02-14
Rpa assistant UI improvements

xylix:
- Change assistant default title to Assisstant from Dialog
- Add some margin for buttons
- Make Run Dialog clear elements and results when it ends by default, earlier documentation was misleading and incorrect (Clear Dialog was claiming that Run Dialog clears when it is started)
- Let Robot Framework auto name keywords in Title Case, by removing unnecessary manual `@keyword("name")` specifications
- Implement a slider element


bogdancondurache (approved):
Looks great!

cmin764 (commented):
LGTM as well!

Seems fine to clear by default when running `Run Dialog`.

PR #830, opened 2023-02-14
Assistant: Add support for default value inputs

bogdancondurache:
Implements https://github.com/robocorp/rpaframework/issues/838

* Adds support for `default` on text inputs
* Adds support for `required` on text inputs
* Makes sure the value of text inputs is always `None` if the field is empty, instead of empty string
* Some docs fixes
* release notes update
* removes `None` values from the results dict


xylix (changes requested):
It’s constantly logging the string when text editing, and also the password field in the test is causing
Exception in thread Thread-40 (change_listener):
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/opt/homebrew/Cellar/python@3.10/3.10.9/Frameworks/Python.framework/Versions/3.10/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/kerkko/Code/rpaframework/packages/assistant/src/RPA/Assistant/flet_client.py", line 174, in change_listener
    handler(e)
TypeError: 'NoneType' object is not callable
Errors when it’s edited. Will take a look.

packages/assistant/src/RPA/Assistant/flet_client.py
```diff
@@ -169,20 +169,24 @@ def on_session_created(conn, session_data):
             # pylint: disable=raise-missing-from
             raise TimeoutException("Reached timeout while waiting for Assistant Dialog")
 
-    def _make_flet_event_handler(self, name: str):
+    def _make_flet_event_handler(self, name: str, handler: Optional[Callable] = None):
         def change_listener(e: ControlEvent):
+            handler(e)
             self.results[name] = e.data
-            e.page.update()
```
xylix, line 175: Need to test that removing this doesn't break anything

packages/assistant/src/RPA/Assistant/flet_client.py
```diff
@@ -169,20 +169,36 @@ def on_session_created(conn, session_data):
...
+        if not handler:
+
+            def change_listener(e: ControlEvent):
+                self.results[name] = e.data
+
+        else:
+
+            def change_listener(e: ControlEvent):
+                handler(e)
+                self.results[name] = e.data
```
bogdancondurache, line 188:
```suggestion
            def change_listener(e: ControlEvent):
                if handler:
                    handler(e)
                self.results[name] = e.data
```

xylix: If you do this one then also remove the comment about why I did the earlier solution 😅 .  (Lines 177-178:

```
# We don't want the if inside the change_listener so it doesn't have to run on
# every on_change event
```

packages/assistant/src/RPA/Assistant/library.py
```diff
@@ -500,6 +502,12 @@ def add_text_input(
         to add a descriptive label and the ``placholder`` text can be given
         to act as an example of the input value.
 
+        The `default` value will be assigned to the input field if the user
+        doesn't complete it. If provided, the placeholder won't be shown.
+        By default it will be an `None`. Also, if a default value is provided
```
cmin764, line 507:
```suggestion
        This is `None` by default. Also, if a default value is provided
```

cmin764 (changes requested):
The static code analysis looks decent!

But when I tested it with this [PR](https://github.com/robocorp/example-signature-match-assistant/pull/4), it looks like the defaults are set, but not provided in the result dictionary:

Screenshot 2023-02-15 at 15 44 51
```
{'submit': 'Submit'}
```
**Solution**: The presence of a `default` should somehow ensure that the internal hook is triggered so that value makes it up to the result dictionary. (as we'll see it only when we focus and interact with the element explicitly)

If we build an additional wrapper/handler that will strip all `None` values from the final dictionary, do you think it will improve the experience?
As instead of getting empty strings, now I get `None`s and I still need to additionally check the value before converting it to a number. Plus the fact that behaviour is still inconsistent: if you leave the element alone, you don't get it in the dictionary, but if you interact with it and leave it empty, you get now a `None`. (in the end the element looks the same in the UI, but you get different results in the output dict if you touch it or not)

----

Might be a good habit to increase the testing coverage when adding a new feature, as it saves reviewing time and bubbles up immediately these kind of issues. (robot functional tests should be enough)
----

docs/source/releasenotes.rst
```diff
@@ -5,6 +5,9 @@ Release notes
 `Upcoming release <https://github.com/robocorp/rpaframework/projects/3#column-16713994>`_
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
+- Library RPA.Assistant:
+    - Bugfixes for dialog clearing
+    - ``Add Slider`` keyword to create sliders for numeric inputs
```
cmin764, line 10:
```suggestion
- Library **RPA.Assistant** (:pr:`830`; ``rpaframework-assistant`` **1.1.0**):
    - Bugfixes for dialog clearing.
    - New ``Add Slider`` keyword to create sliders for numeric inputs.
```
And is better to use instead of `:pr:` directive the issue one:
```
:issue:`<nr>`
```
Once you create and link an Issue to this PR. (so we can track them down in the release notes -- see examples from past releases below)

cmin764 (commented):
Left a nitpick on formatting the release notes.

packages/assistant/src/RPA/Assistant/flet_client.py
bogdancondurache: tested, I don't think it was required to begin with, because we don't need to re-render anything on-change

packages/assistant/src/RPA/Assistant/library.py
bogdancondurache: thanks! writing docs is hard 😅  changed to your suggestion

bogdancondurache:
> The static code analysis looks decent!
> 
> 1. But when I tested it with this [PR](https://github.com/robocorp/example-signature-match-assistant/pull/4), it looks like the defaults are set, but not provided in the result dictionary:
>    Screenshot 2023-02-15 at 15 44 51
>    ```
>    {'submit': 'Submit'}
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    **Solution**: The presence of a `default` should somehow ensure that the internal hook is triggered so that value makes it
>    up to the result dictionary. (as we'll see it only when we focus and interact with the element explicitly)
> 2. If we build an additional wrapper/handler that will strip all `None` values from the final dictionary, do you think it will improve the experience?
>    As instead of getting empty strings, now I get `None`s and I still need to additionally check the value before converting it to a number. Plus the fact that behaviour is still inconsistent: if you leave the element alone, you don't get it in the dictionary, but if you interact with it and leave it empty, you get now a `None`. (in the end the element looks the same in the UI, but you get different results in the output dict if you touch it or not)
> 
> Might be a good habit to increase the testing coverage when adding a new feature, as it saves reviewing time and bubbles up immediately these kinds of issues. (manual robot functional tests should be enough)

Both really good points, fixed 1. and implemented 2. Thanks for the great suggestion

docs/source/releasenotes.rst
bogdancondurache: great points, thanks! fixed

xylix:
@cmin764 I made small bugfixes and refactored the code a bit to make validations and required fields work better together, if you want to take one more look feel free, otherwise this should be getting ready for merge.

PR #834, opened 2023-02-15
Add some error catching to windows recorder, print in console that ESC stops recording

xylix:
I could take a short look in actually implementing the keyboard recording, but just these changes should make the recorder a bit easier to use. 

Reasoning for the fixes:

- Printing "ESC stops recording": I was originally trying to stop it with ctrl-c and wondering why it didn't record, before I managed to read the actual docs. Actual users might come across same problem.
-  changed listening started print from "keyboard and mouse listeners started" to "Mouse recording started. Use ESC to stop recording." to be more clear to users that currently only mouse is being recorded.
- Add error catching to make traces like in first screenshot
Screenshot 2023-02-15 at 11 40 31

```log
(.venv) PS C:\Users\kerkko\Downloads\rpafw-windows> python -m RPA.Windows
Keyboard and mouse listeners started. Use ESC to stop recording.
Could not capture element, got exception (-2147024809, 'The parameter is incorrect.', (None, None, None, 0, None))
recording continues
Could not capture element, got exception (-2147024809, 'The parameter is incorrect.', (None, None, None, 0, None))
recording continues
Could not capture element, got exception (-2147024809, 'The parameter is incorrect.', (None, None, None, 0, None))
recording continues
Could not capture element, got exception (-2147024809, 'The parameter is incorrect.', (None, None, None, 0, None))
recording continues

--------------------------------------------------------------------------------
COPY & PASTE BELOW CODE INTO *** Tasks *** or *** Keywords ***
--------------------------------------------------------------------------------

Control Window    Program Manager  # Handle: 65794
Click   name:Desktop

--------------------------------------------------------------------------------
```

Couple of questions:
- how difficult would implementing the keyboard recording be?
- there seems to be a bug that when you click on an element that causes the 'The parameter is incorrect.' type of exceptions, the rest of recording doesn't work. (Both before and after my try catch.) I will try to fix this, but ideas why this happens are welcome. (Is either Pynput or the `from RPA.core.windows.inspect import ElementInspector, RecordElement` functionality we use somehow stateful, if wrong Window is initially selected then future recordings will just fail?)

### DIFF
--- a/packages/windows/src/RPA/Windows/main.py
+++ b/packages/windows/src/RPA/Windows/main.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 from typing import List
 
+from comtypes import COMError
 import fire
 from pynput_robocorp import mouse, keyboard  # pylint: disable=C0415
 from RPA.core.windows.inspect import ElementInspector, RecordElement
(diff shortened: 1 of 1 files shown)


mikahanninen (approved):
LGTM

PR #855, opened 2023-02-23
Windows improvements: `Print Tree`, `Set Mouse Movement`, `Set Value`

### LINKED ISSUE #483: RPA.Windows add more logging and error reporting into `Set Value` keyword
additions
- [x] add info level logging on which action keyword executed
- [x] check at the end if the value was actually changed - error if not
- [x] add option that value is changed with `Send Keys` as the last option

### LINKED ISSUE #591: `RPA.Windows` problem recognizing elements with different than normal display scaling
Even when clicking with locators (no coordinates), if the scaling is different from 100% we might encounter issues on recognizing and clicking elements with `RPA.Windows`.
The issue was solved by setting to 100% normal scaling. @mikahanninen please remove the `bug` tag from this one if this consists a limitation that has to be documented only.

- Forum [post](https://forum.robocorp.com/t/click-method-not-working/1948/4?u=cosmin)

### LINKED ISSUE #791: RPA.Windows select action keyword simulates mouse movement by default
The select keyword in action module calls the windows automation library with default mouse movement simulation turned on (simulateMove=True). This does not work for our application, it is way too slow and causes many wrong combobox selections.

Selection works if we turn off mouse movement simulation by calling instead:
element.item.Select(value, simulateMove=False)

I would think that turning mouse simulation off is a better default setting, because it is faster and more reliable.
It may even be helpful to have "simulateMove" as an API parameter so that the test author can decide whether they want to move the mouse or not.

### LINKED ISSUE #796: RPA.Windows Can not find element with id
I try to use the open source application "KeePass" to write all groups with there entries into a file. I have been successful in wlaking thru all groups (Keyword Get all entries of the group) but in Keyword "List Group" I fail to find the Element "m_lvEntries"..
If I create a new robot file using the cding of keyword  "List Group" as new task and copy all my created keywords position manually to one of the groups and start the robot then the output of the entries is successful.

Can you pleae help me and tell me how to find the element "m_lvEntries" in my keyword "List Group".

My coding is the file ListG.txt (a copy of ListG.robot)
The output.txt is a copy of output.xml
The log.txt is a copy of log.html

[ListG.txt](https://github.com/robocorp/rpaframework/files/10555810/ListG.txt)

[output .txt](https://github.com/robocorp/rpaframework/files/10555816/output.txt)

[log.txt](https://github.com/robocorp/rpaframework/files/10555826/log.txt)

### LINKED ISSUE #797: `RPA.Windows.Control Window` problem on older Windows versions
**description of the issue**
The reported occurrence happens on Windows Server 2012 where `Control Window` just activates the application icon in the Task Bar but does not get control of the window nor bring the window to the foreground.

**analysis**
The control of the window can be achieved by an alternate way of setting the window for the library instance via accessing elements returned by `List Windows` keyword. 

requires still more investigation

cmin764:
Closes #796 
Closes #791 
Closes #483
Closes #591
Closes #797 

### Fixes
- `Print Tree`:
  - Correct depth (starts from 0 because it starts with the **root** and not with the children first)
  - Partial locator paths visible in the returned structure (so you can copy-paste that to get the element using the `path:` strategy)
- Mouse movement simulation switching keyword
- `Set Value` enhancements
- Docs on display scaling issue

### ToDo
- [x] Test mouse move simulation switch with: `Click`, `Select`
- [x] Python example docstring for `Set Value`
- [x] Test EOLs passed with `Set Value` in different scenarios (enable switches, keys sending etc.)
- [x] Research `TextChildPattern` for obtaining a `TextRange` which can check the set value when using keys sending
- [x] Docs: Explain depth vs. element position (by index) in the printed tree
- [x] Add @mikahanninen Portal example on `Print Tree` usage in the keyword docstring

### Next
- Add release notes and publish with the upcoming **main**.
- Test it against Windows 10/11 and Server.


cmin764:
Given the `Path Explore Notepad` test [bot](https://github.com/cmin764/robots/blob/main/desktop-auto/tasks.robot#L305-L329), the tree of elements is displayed and returned correctly now:
Screenshot 2023-02-23 at 18 08 39
Screenshot 2023-02-23 at 18 09 03

mikahanninen (approved):

PR #860, opened 2023-02-28
`RPA.PDF` Add `boxes_flow` to available conversion settings

mikahanninen:
The `boxes_flow` affects how text boxes contained in the PDF page are ordered. In some cases the default setting will not return text boxes in consistent order even when PDF's are using same PDF template but different data.

The acceptable values for the new setting are between float range `-1.0...+1.0` OR `None` which disables this feature in the **pdfminer**.

```robotframework
Set Convert Settings    boxes_flow=${NONE}
```


packages/pdf/src/RPA/PDF/keywords/model.py
```diff
@@ -478,8 +479,8 @@ class ModelKeywords(LibraryContext):
     @keyword
     def convert(
         self,
-        source_path: str = None,
-        trim: bool = True,
+        source_path: Optional[str] = None,
+        trim: Optional[bool] = True,
```
cmin764, line 483: Nit: Actually this should stay: `trim: bool = True,` as you usually pass `True`/`False` to the param, not `None`.

packages/pdf/src/RPA/PDF/keywords/model.py
```diff
@@ -581,8 +582,8 @@ def _decode_field(
     def get_input_fields(
         self,
         source_path: Optional[str] = None,
-        replace_none_value: bool = False,
-        encoding: str = FIELDS_ENCODING,
+        replace_none_value: Optional[bool] = False,
```
cmin764, line 585: Same as above on line **483**.

packages/pdf/src/RPA/PDF/keywords/model.py
```diff
@@ -32,6 +32,7 @@
 
 
 Coords = Tuple[int, ...]
+BOXES_FLOW_NOT_SET = -999.99
```
cmin764, line 35: Would be best to set it to `0.5` as [default](https://pdfminersix.readthedocs.io/en/latest/reference/composable.html#laparams), so we'd keep the behavior consistent. (and set this value no matter what inside as param -> the effect stays the same if the user doesn't provide a different value)

cmin764 (commented):
### Post-merge review

Late at the party, still I can take care of these comments in my next [Issue](https://github.com/robocorp/rpaframework/issues/872).

PR #861, opened 2023-02-28
RPA.Assistant v2.0.0 flet 0.4.2 

xylix:
Current drafted release notes:

RPA.Assistant (``rpaframework-assistant`` **2.0.0**):
    -  Breaking  changes:
        - Run Dialog and Ask User parameter "clear" was removed, clearing is now mandatory
        - (affects Python side users only): Location enum that ``run_dialog`` and 
        ``ask_user`` accepts was changed from Location to WindowLocation to improve clarity. 

    - Update to flet 0.4.2, library startup time is improved from ~1.2-2 seconds to 0.2-1 second
    - Added location argument to Add Button keyword


docs/source/releasenotes.rst
```diff
@@ -5,7 +5,14 @@ Release notes
...
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
-**RPA.Assistant** (``rpaframework-assistant`` **1.2.X**):
+**RPA.Assistant** (``rpaframework-assistant`` **2.0.0**):
+    - ** Breaking ** changes:
+        - Run Dialog and Ask User parameter "clear" was removed, clearing is now mandatory
+        - (affects Python side users only): Location enum that ``run_dialog`` and 
+        ``ask_user`` accepts was changed from Location to WindowLocation to improve clarity. 
+
+    - Update to flet 0.4.2, library startup time is improved from ~1.2-2 seconds to 0.2-1 second
```
bogdancondurache, line 14: this is actually misleading because we were pre-loading before; if we keep this in the change notes, users will expect huge improvements when it won't really be the case

bogdancondurache (commented):
looks good, one small comment on the change notes

docs/source/releasenotes.rst
xylix: mm right only the cold startup time is improved greatly. I guess then we could leave this out. I will look if there are some relevant bugfixes / other stuff from flet 0.4.2.

bogdancondurache (approved):

PR #862, opened 2023-02-28
Update the docs to mention the deprecation

bogdancondurache:


packages/dialogs/src/RPA/Dialogs/library.py
```diff
@@ -16,7 +16,12 @@
 
 @library(scope="GLOBAL", doc_format="REST", auto_keywords=False)
 class Dialogs:
-    """The `Dialogs` library provides a way to display information to a user
+    """**NOTICE:** The `Dialogs` library is now marked for deprecation.
+    The library will be replaced by `Assistant`. In order to get more
+    information and check the changes and the migration guide, please
```
cmin764, line 21:
```suggestion
    information about the changes and the migration guide, please
```
to avoid doubling "and" & "check".

packages/dialogs/src/RPA/Dialogs/library.py
```diff
@@ -16,7 +16,12 @@
...
+    check the Release Notes: https://rpaframework.org/releasenotes.html#id4
```
cmin764, line 22:
```suggestion
    check the Release Notes: https://updates.robocorp.com/release/Gfko5-rpaframework-2200
```
or
```suggestion
    check the release updates: https://updates.robocorp.com/
```

Beware that the .org release notes aren't stable and they may change links when you add a new one or change the file.

cmin764 (changes requested):
LGTM so far, apart from the listed changes, here is one more improvement that can be brought:

Also adding a specific deprecation warning message appearing in the logs when such library is used. ([example](https://github.com/robocorp/rpaframework/blob/master/packages/main/src/RPA/Desktop/Windows.py#L216-L220) with `RPA.Desktop.Windows`)

As not even us the engineers don't check docs frequently for such updates, might be beneficial to see right in the running bot when we advise to migrate to "the new black".

packages/dialogs/src/RPA/Dialogs/library.py
bogdancondurache: Oh, interesting, Kari requested this URL, but I guess he also didn't knew it can change, thank you!

bogdancondurache:
> LGTM so far, apart from the listed changes, here is one more improvement that can be brought:
> 
> Also adding a specific deprecation warning message appearing in the logs when such library is used. ([example](https://github.com/robocorp/rpaframework/blob/master/packages/main/src/RPA/Desktop/Windows.py#L216-L220) with `RPA.Desktop.Windows`)
> 
> As not even us the engineers usually check docs frequently for such updates, so might be beneficial to see right in the running bot when we advise to migrate to "the new black".

Thanks, had no idea about the deprecation function, added it now as well!

cmin764 (approved):

PR #863, opened 2023-03-01
Relax rpaframework-core packaging dependency from 21^ to >=21, <=24

xylix:
Currently rpaframework-core doesn't support newer `packaging` than 21. This relaxes that dependency.

Did not see any changes in https://packaging.pypa.io/en/latest/changelog.html that would break rpaframework-core's configuration because of this.

Motivation:

the packaging=^21.3 dependency in rpaframework-core was making the new Flet 0.4.2 install together with -core impossible. So I relaxed the packaging dependency to packaging>=21.3,<24 . (Packaging version numbers go by “release year”, and they do contain some breaking changes but I didn’t see anything that would affect us.)


cmin764 (commented):
Post-merge: LGTM!

PR #870, opened 2023-03-07
RPA.Database : Fix psycopg2 module error

aabashkin:
The following error occurs in RPA.Database while calling a stored procedure using the psycopg2 module:

`[FAIL] TypeError: 'as_dict' is an invalid keyword argument for this function`

Several modules ("cx_Oracle", "oracledb") are already excluded from this. This PR simply adds psycopg2 to the list.

For reference my dependencies are:
rpaframework==22.0.0
psycopg2-binary==2.9.5

### DIFF
--- a/packages/main/src/RPA/Database.py
+++ b/packages/main/src/RPA/Database.py
@@ -410,7 +410,7 @@ class Database:
         params = params or []
         cur = None
         try:
-            if self.db_api_module_name in ("cx_Oracle", "oracledb"):
+            if self.db_api_module_name in ("cx_Oracle", "oracledb", "psycopg2"):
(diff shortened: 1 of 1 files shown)


mikahanninen:
@aabashkin Note about calling a stored procedure. I tested using Postgresql15 and the keyword ``Call Stored Procedure`` did not work in my example test case.

This is what worked. I had to set ``autocommit=${TRUE}`` and use ``Query`` keyword.
```robotframework
Connect To Database    psycopg2    ${DATABASE}    ${USER}    ${PASSWORD}    127.0.0.1    autocommit=${TRUE}
${result}=    Query    call transfer(1,2,600.00) 
```
This will be included in the release I will make later today.

PR #873, opened 2023-03-07
RPA.Assistant: add layout support

xylix:
**RPA.Assistant** (``rpaframework-assistant`` **2.1.0**):
    - Add ``Set Title`` keyword. Can be used to set title of assistant when it is running.

    - Add ``Open Row`` and ``Close Row`` keywords. Used to layout elements into rows.
    - Add ``Open Column`` and ``Close Column`` keywords. Used to layout elements into columns.
    - Add ``Open Navbar`` and ``Close Navbar`` keywords. Can be used to create an always visible top bar for a dialog.
    
    - Add ``Open Stack`` and ``Close Stack`` keywords. Can be used to position elements manually.
    - Add ``Open Container`` and ``Close Container`` keywords. Useful for styling or placing single elements.

    - Add ``Add Flet Icon`` that enables adding any icons from flets large gallery of icons
        - Compared to `Add Icon` it is more difficult to use but supports a much larger amount of icons.


packages/assistant/src/RPA/Assistant/background_flet.py
```diff
@@ -12,6 +13,7 @@ class BackgroundFlet:
     """Class that manages the graphical flet subrocess and related operations"""
 
     def __init__(self):
+        atexit.register(self.close_flet_view)
```
bogdancondurache, line 16: this seems to be missing from the described changes, is this a regression (aka something we've lost along the way) or is it something we thought it's not needed when it actually is?

packages/assistant/src/RPA/Assistant/flet_client.py
```diff
@@ -29,11 +36,13 @@ class TimeoutException(RuntimeError):
     """Timeout while waiting for dialog to finish."""
 
 
-class Elements(NamedTuple):
+@dataclass
```
bogdancondurache, line 39: great usage of a dataclass 👍

packages/assistant/src/RPA/Assistant/flet_client.py
```diff
@@ -137,8 +147,30 @@ def add_element(
...
         # make a container that adds margin around the element
-        container = Container(margin=5, content=element)
-        self._elements.visible.append(container)
+
+        # if it's a Container we don't create our own margin container to allow setting
+        # of padding and absolute position manually by user
+        if isinstance(element, Container):
+            new_element = element
+        else:
+            new_element = Container(margin=5, content=element)
```
bogdancondurache, line 156: maybe good to document that all elements will have a margin of 5 unless the user creates the containers himself?

packages/assistant/src/RPA/Assistant/flet_client.py
```diff
@@ -137,8 +147,30 @@ def add_element(
...
+
+        if self._layout_stack != []:
+            current_layout = self._layout_stack[-1]
+            if hasattr(current_layout, "controls"):
+                # Row, Stack and Column have "controls"
+                current_layout.controls.append(new_element)
+            elif hasattr(current_layout, "actions"):
```
bogdancondurache, line 163: wouldn't `isinstance` be safer here?

packages/assistant/src/RPA/Assistant/background_flet.py
xylix: Yeah. If running through pure python and using the red-x on macOS to quite the application, the flet UI stays around. 

This fixes that.

packages/assistant/src/RPA/Assistant/flet_client.py
xylix: Yeah that's already been the behaviour since 1.<something>, I thought I documented it in the release notes but cannot find it.

I guess the question is that should we document this around the Open Container keyword or at the top level library documentation? I was actually considering adding a sub-heading into the docs like what playwright lib does here:
Screenshot 2023-03-08 at 10 21 51

talking about layouting in general. That would fix this issue.

packages/assistant/src/RPA/Assistant/flet_client.py
xylix: True. Actually we should do both `hasattr` and `isinstance` check, so that our tests will easily catch if the Flet API regarding these classes has an update that removes these properties.

packages/assistant/src/RPA/Assistant/background_flet.py
xylix: Documented to releasenotes.

packages/assistant/src/RPA/Assistant/flet_client.py
xylix: Thanks!

packages/assistant/src/RPA/Assistant/flet_client.py
xylix: Documented in 0525e0dc

packages/assistant/src/RPA/Assistant/flet_client.py
xylix: fixed to isinstance.

PR #887, opened 2023-03-14
CodeQL warning fixes

xylix:
Code QL is currently giving a couple errors when it's run: see https://github.com/robocorp/rpaframework/actions/workflows/codeql-analysis.yml

```
Warning: 1 issue was detected with this workflow: Please specify an on.push hook so that Code Scanning can compare pull requests against the state of the base branch.
Setup CodeQL tools
Load language configuration
Setup Python dependencies
Warning: An error occurred while trying to automatically install Python dependencies: Error: The process '/usr/bin/python3' failed with exit code 1
```

This PR attempts fixes at these. Adds an on-push trigger for the workflow on pushes to master, and manual installation of the dependencies.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -56,7 +56,7 @@ jobs:
         poetry install
         # Set the `CODEQL-PYTHON` environment variable to the Python executable
         # that includes the dependencies
-        echo "CODEQL_PYTHON=$(which python)" >> $GITHUB_ENV
+        echo "CODEQL_PYTHON=$(poetry env info -p)" >> $GITHUB_ENV
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -15,6 +15,8 @@ on:
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ "master" ]
+  push:
+    branches: [ "master" ]
   schedule:
     # https://crontab.guru/#23_21_*_*_0
     # "At 21:23 on Sunday."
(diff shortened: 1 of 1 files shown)


xylix:
Another thing to note:

We are analyzing a large amount of code, when inlcuding deps:
```
|             Diagnostic              |   Summary   |
+-------------------------------------+-------------+
| Successfully extracted Python files | 229 results |

Analysis produced the following metric data:

|                         Metric                          |  Value  |
+---------------------------------------------------------+---------+
| Total lines of user written Python code in the database |   56033 |
| Total lines of Python code in the database              | 1398075 |


/opt/hostedtoolcache/CodeQL/2.12.3-20230217/x64/codeql/codeql database print-baseline /home/runner/work/_temp/codeql_databases/python
Counted a baseline of 57539 lines of code for python.
```

The HW recommendations for github codeQL are like this:
https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/recommended-hardware-resources-for-running-codeql
Screenshot 2023-03-14 at 14 09 29

This probably explains why the CodeQL analysis takes ~15 minutes to run.

.github/workflows/codeql-analysis.yml
```diff
@@ -39,6 +41,22 @@ jobs:
...
+      run: |
+        python -m pip install --upgrade pip
+        pip install poetry
+
+    - name: Install dependencies
+      run: |
+        poetry install
+        # Set the `CODEQL-PYTHON` environment variable to the Python executable
+        # that includes the dependencies
+        echo "CODEQL_PYTHON=$(which python)" >> $GITHUB_ENV
```
cmin764, line 59:
```suggestion
        echo "CODEQL_PYTHON=$(poetry env info -p)" >> $GITHUB_ENV
```
as by judging:
>         # Set the `CODEQL-PYTHON` environment variable to the Python executable
>         # that includes the dependencies

It looks like we're aiming for the deps installed by Poetry, which usually creates a venv with these isolated from the default system/active `python`. Or maybe I misunderstood the comment, but then I don't understand why do we issue before a `poetry install` command.

cmin764 (approved):
Overall LGTM, I just have a doubt about the Python interpreter setting.

Btw, will this also increase speed by analyzing just the code we write or are we ok with the gate finishing after 15min after analyzing all the packages in the Poetry venv?

.github/workflows/codeql-analysis.yml
xylix: I somehow thought poetry would install globally here but yeah you are correct, your suggestion should be way more correct.

.github/workflows/codeql-analysis.yml
```diff
@@ -56,7 +56,7 @@ jobs:
         poetry install
         # Set the `CODEQL-PYTHON` environment variable to the Python executable
         # that includes the dependencies
-        echo "CODEQL_PYTHON=$(which python)" >> $GITHUB_ENV
+        echo "CODEQL_PYTHON=$(poetry env info -p)" >> $GITHUB_ENV
```
cmin764, line 59: Sorry: `$(poetry env info -p)/bin/python`

cmin764 (approved):
I figured out I recommended a bad command.

PR #899, opened 2023-03-22
Feature: support parameterized sql queries in RPA.Database

aabashkin:
## What?
Added support for parameterized SQL queries to the Query keyword in the Database module

## Why?
SQL queries are susceptible to SQL injection vulnerabilities using standard string concatenation/formatting techniques. Utilizing the parameterization abilities of the Python DB API will allow users to safely create queries that contain untrusted input.

## How?
Added a 'data' parameter to the Query keyword, which is then passed to 'cursor.execute'

## Testing?
New Python and robot tests have been added

## Anything Else?
Example in the documentation has been updated with the parameterized query version. This is in accordance with the 'secure by default' principle.
Got rid of 'RETURNING' portion of SQL statement both in the documentation and the tests because it fails in sqlite, which is what is being used in this case. Replaced by an additional SELECT. This should make the example more functional.
The following resource was referenced during development: [https://bobby-tables.com/python](https://bobby-tables.com/python)


aabashkin:
@mikahanninen any idea what the issue could be with the failing tests? Seems like a build environment issue.

aabashkin:
@mikahanninen you were right, all green now 👍

PR #900, opened 2023-03-22
Relax core pillow dep from 9.1.1 -> 9.0.1

xylix:
Enables installing rpaframework-core and packages that depend on it together with packages that depend on pillow 9.0.x, like `label-studio`


xylix:
closed for now, another approach will be first attempted with the customer project

Issue #454, opened 2022-04-05
RPA.Email.ImapSmtp Send Message should not require recipients


antusystem:
Hi, after adding **BCC** and **CC** to **Send Message** I don't think that **recipients** should be a requirement because you could send a mail to 100 people but in the field **BCC** so they don't see each others email. Right now, I would have to add a recipient and add in **BCC** all the other mails. The code should work like this:

```
${secret}=    Get Secret    MySecrets
Authorize    account=${secret}[username]    password=${secret}[password]
Send Message    sender=${secret}[username]
# ...    recipients=mail1@domain.com
...    bcc=mail2@domain.com,mail3@domain.com,mail4@domain.com
...    subject=My Subject
...    body=My Body
```
If you try to execute it like that the errors is: `Keyword 'RPA.Email.ImapSmtp.Send Message' missing value for argument 'recipients'.`

Later I could check the library to see if I can colaborate, but right now I don't know how it is working (and also I'm a little busy).

Adstefnum:
Hello, @cmin764 I would like to work on this project and once I pass it I will work on the other RPA.Email.ImapSmtp improvements. I have forked the repo and am working on it. Let me know if I can go on.

PR #930, opened 2023-04-23
Close RPA.Email.ImapSmtp Send Message should not require recipients #454

Adstefnum:
I gave the recipients a default value of an empty list so it is not required.

I checked if the recipients is empty after everything e.g cc and bcc have been added to it.


mikahanninen:
Thank you for the pull request 👍 I will look into this for the next release.

mikahanninen:
**The fix** - `Send Message` keyword can be called by giving one of the following: `recipients`, `cc`, or `bcc`. If not given then `NoRecipientsError` exception is thrown by the keyword.

PR #950, opened 2023-05-09
Upgrade to Robot Framework 6

mikahanninen:
packages requiring upgrade before `rpaframework` can be upgraded

- [x]  robotframework-docgen [[repo](https://github.com/robocorp/robotframework-docgen)] == release 0.16.0
- [ ]  rpaframework-pdf
- [ ]  rpaframework-windows

**main** upgrade

- [ ]  rpaframework

other package upgrades

- [ ]  rpaframework-assistant
- [ ]  rpaframework-aws
- [ ]  rpaframework-google
- [ ]  rpaframework-openai (due to the robotframework-docgen)
- [ ]  rpaframework-dialogs (MAYBE NOT, deprecated)

NO package upgrade needed

- rpaframework-recognition

**IMPROVEMENTS**
- RF 6.0 "Enhancements to automatic argument conversion"
   - Argument types have better automatic conversion and validation
- RF 6.0 "start/end_keyword listener methods get more information about control structures"
   -  Possible to create better listeners
- RF 6.1 "Possibility to flatten keyword structures during execution"
   - Possible to flatten keywords using keyword tag "robot:flatten"
- RF 6.1 "JSON variable file support"
   - Possible to create variables from JSON file (YAML support has existed already), for example. "Variables        example.json"
- RF 6.1 "WHILE loop enhancements"
   - Adds optional "on_limit" which can be given what to do when limit is reached (instead of current default FAIL)
   - Loop condition is optional
   - Adds optional "on_limit_message"
 - RF 6.1 "New pseudo log level CONSOLE"
    - Effectively logs a message as INFO level and also to the console
    
**CHANGES TO PREVIOUS BEHAVIOUR (deprecations/breaking changes)**
- RF 6.0 "Enhancements to keyword namespaces"
   - Keyword with the same name earlier caused conflict if keyword exist in another resource file. Now the keyword will be used.
- RF 6.0 "Enhancements for setting keyword and test tags"
   - "Force Tags" --> "Task Tags", "Default Tags" --> "Tags"
- RF 6.0 "WITH NAME in favor of AS when giving alias to imported library"
   - Library import aliases new token is "AS"
- RF 6.1 "Changes to output.xml"
- RF 6.1 "Python 3.7 support"


packages/main/pyproject.toml
```diff
@@ -39,7 +39,7 @@ rpaframework-core = "^11.0.1"
 rpaframework-pdf = "^7.1.5"
 rpaframework-windows = { version = "^7.3.2", platform = "win32" }
 jsonpath-ng = "^1.5.2"
-robotframework = ">=4.0.0,!=4.0.1,<6.0.0"
+robotframework = ">=4.0.0,!=4.0.1,<7.0.0"
```
custompointofview, line 42: Hey... just jumping in after the meeting. 😊 
If we'd like to avoid `6.1.0` should these clauses be more restrictive?

mikahanninen: Yeah I will add 6.1.0 exclusion.

PR #955, opened 2023-05-19
Do NOT merge: testing non-async (Selenium related)

cmin764:
As failures were happening here: https://github.com/robocorp/rpaframework/pull/943 (tests running async)


cmin764:
Since all the pipelines are running just fine, I'm going to close it. But re-open it for testing when there another flaky robot test case will appear again.

PR #980, opened 2023-06-02
Lock `O365` to Python 3.7 compatible version

blakewaud:
v2.0.27 of O365 includes a f-string that is not compatible with Python 3.7


packages/windows/pyproject.toml
```diff
@@ -70,8 +70,8 @@ windows-record = 'RPA.scripts.record:main'
 target-version = ["py37", "py38", "py39"]
 
 [build-system]
-requires = ["poetry>=0.12"]
-build-backend = "poetry.masonry.api"
+requires = ["poetry-core>=1.0.0"]
+build-backend = "poetry.core.masonry.api"
```
cmin764, line 74: Good catch for the alignment!

cmin764 (approved):
LGTM! And luckily, you'd have to release **main** only, as the lock-files of all the other sub-packages aren't affecting the final published package. (as for quite some time, we ship within that the _pyproject.toml_ file instead of a _setup.py/cfg_ script)

cmin764:
Side-challenge: Do you know any way of outputting the [failed](https://github.com/robocorp/rpaframework/actions/runs/5180042853/jobs/9337980585?pr=980) test(s) when running in async (`-a`) mode? (this will aid a LOT the DX)

blakewaud:
@cmin764 the underlying async behavior is provided by `invoke`. I'm copying the associated API documentation here for discussion because as far as I can tell, it _should_ be getting joined to std out.

> Connections to the controlling terminal are disabled, meaning you will not see the subprocess output and it will not respond to your keyboard input - similar to hide=True and in_stream=False (though explicitly given (out|err|in)_stream file-like objects will still be honored as normal).
> 
> [run](https://docs.pyinvoke.org/en/stable/api/runners.html#invoke.runners.Runner.run) returns immediately after starting the subprocess, and its return value becomes an instance of [Promise](https://docs.pyinvoke.org/en/stable/api/runners.html#invoke.runners.Promise) instead of [Result](https://docs.pyinvoke.org/en/stable/api/runners.html#invoke.runners.Result).
> 
> [Promise](https://docs.pyinvoke.org/en/stable/api/runners.html#invoke.runners.Promise) objects are primarily useful for their [join](https://docs.pyinvoke.org/en/stable/api/runners.html#invoke.runners.Promise.join) method, which blocks until the subprocess exits (similar to threading APIs) and either returns a final [Result](https://docs.pyinvoke.org/en/stable/api/runners.html#invoke.runners.Result) or raises an exception, just as a synchronous run would.
> 
> As with threading and similar APIs, users of asynchronous=True should make sure to join their [Promise](https://docs.pyinvoke.org/en/stable/api/runners.html#invoke.runners.Promise) objects to prevent issues with interpreter shutdown.
> 
> One easy way to handle such cleanup is to use the [Promise](https://docs.pyinvoke.org/en/stable/api/runners.html#invoke.runners.Promise) as a context manager - it will automatically join at the exit of the context block.

And the result object's documentation describes how the streams are returned:

> A container for information about the result of a command execution.
> 
> All params are exposed as attributes of the same name and type.
> 
> Parameters
> stdout ([str](https://docs.python.org/2.7/library/functions.html#str)) – The subprocess’ standard output.
> 
> stderr ([str](https://docs.python.org/2.7/library/functions.html#str)) – Same as stdout but containing standard error (unless the process was invoked via a pty, in which case it will be empty; see [Runner.run](https://docs.pyinvoke.org/en/stable/api/runners.html#invoke.runners.Runner.run).)
> 
> encoding ([str](https://docs.python.org/2.7/library/functions.html#str)) – The string encoding used by the local shell environment.
> 
> command ([str](https://docs.python.org/2.7/library/functions.html#str)) – The command which was executed.
> 
> shell ([str](https://docs.python.org/2.7/library/functions.html#str)) – The shell binary used for execution.
> 
> env ([dict](https://docs.python.org/2.7/library/stdtypes.html#dict)) – The shell environment used for execution. (Default is the empty dict, {}, not None as displayed in the signature.)
> 
> exited ([int](https://docs.python.org/2.7/library/functions.html#int)) –
> 
> An integer representing the subprocess’ exit/return code.
> 
> Note
> This may be None in situations where the subprocess did not run to completion, such as when auto-responding failed or a timeout was reached.
> 
> pty ([bool](https://docs.python.org/2.7/library/functions.html#bool)) – A boolean describing whether the subprocess was invoked with a pty or not; see [Runner.run](https://docs.pyinvoke.org/en/stable/api/runners.html#invoke.runners.Runner.run).
> 
> hide (tuple) –
> 
> A tuple of stream names (none, one or both of ('stdout', 'stderr')) which were hidden from the user when the generating command executed; this is a normalized value derived from the hide parameter of [Runner.run](https://docs.pyinvoke.org/en/stable/api/runners.html#invoke.runners.Runner.run).
> 
> For example, run('command', hide='stdout') will yield a [Result](https://docs.pyinvoke.org/en/stable/api/runners.html#invoke.runners.Result) where result.hide == ('stdout',); hide=True or hide='both' results in result.hide == ('stdout', 'stderr'); and hide=False (the default) generates result.hide == () (the empty tuple.)

And finally, our code joins those promises and prints both streams:

> ```python
>     for test, result in results.items():
>         if color:
>             result_header = Fore.BLUE + f"Results from {test} tests:"
>             result_msg = Style.RESET_ALL + remove_blank_lines(result.stdout)
>         else:
>             result_header = f"*** Results from {test} tests:"
>             result_msg = remove_blank_lines(result.stdout)
>         print(result_header)
>         print(result_msg)
>         if hasattr(result, "stderr"):
>             print(remove_blank_lines(result.stderr))
>         print(os.linesep)
> ```

PR #1003, opened 2023-07-07
 `RPA.Salesforce`. Add Salesforce OAuth2 authentication and execute_apex

mikahanninen:
Pull request adds two keywords

- Authenticate With Connected App
- Execute Apex

TODO
- [x] tests done
- [x] code examples
- [x] release note added


packages/main/src/RPA/Salesforce.py
```diff
@@ -265,6 +265,47 @@ def auth_with_token(self, username: str, password: str, api_token: str) -> None:
...
+        response = requests.post(
+            f"https://{self.domain}.salesforce.com/services/oauth2/token",
+            json={
+                "username": username,
+                "password": password + api_token,
+                "client_id": consumer_key,
+                "client_secret": consumer_secret,
+                "grant_type": "password",
+            },
+        ).json()
```
cmin764, line 296: I would run a `.raise_for_status()` over the received response, then turn that into a JSON. (this way we'll have it breaking if an error is received during the request)
Then catch and reraise `SalesforceAuthenticationError(response)` from this one.

packages/main/src/RPA/Salesforce.py
```diff
@@ -265,6 +265,47 @@ def auth_with_token(self, username: str, password: str, api_token: str) -> None:
...
+
+        if "access_token" in response.keys() and response["access_token"]:
```
cmin764, line 298:
```suggestion
        if response.get[("access_token"):
```
might be more clean.

cmin764 (commented):
LGTM, nothing critical, just a few nits.

PR #1016, opened 2023-07-18
Sync with `robocorp-storage` 0.3.*

cmin764:
### ToDo
- [x] Fix logging typo in upstream lib (`Failed to log following message properly: Uploading asset (content-type: %s)`)
- [x] Add release note including migration guide given the breaking change

### Next
- Release this version and announce the other releases as well
- Post-release follow-up GH/Linear issue for env vars pull from CR endpoint


packages/main/src/RPA/Robocorp/Storage.py
```diff
@@ -78,61 +81,229 @@ def print_all_assets():
         return storage.list_assets()
 
     @keyword
-    def get_asset(self, name: str) -> str:
-        """Get an asset's value by providing its `name`.
+    def set_bytes_asset(self, name: str, value: bytes, wait: bool = True):
```
cmin764, line 84: Keeping the interface simple with one distinction only between a `value` (no matter the type) and a file `path`.

packages/main/src/RPA/Robocorp/Storage.py
```diff
@@ -78,61 +81,229 @@ def print_all_assets():
...
+
+        .. code-block:: python
+
+            def set_an_asset():
+                storage.set_file_asset("my-file-asset", "report.pdf")
+        """
+        storage.set_file(name, path=path, wait=wait)
+
+    @keyword
+    def get_file_asset(self, name: str, path: str, overwrite: bool = False) -> str:
```
cmin764, line 273: Usually `overwrite` is more appropriate since we aren't just fine if a file exists with `exist_ok`, because the end result isn't the same, as we might end up with a completely different file content even if the name is the same. (`exist_ok` applies best when making a directory and we're just ok if it already exists, as it doesn't produce an undesired effect @osrjv -- no need to change this in **robo** as well)

packages/main/tests/robot/test_robocorp_storage.robot
cmin764, line 1: Fill in the `api_token` with a real valid token in _secrets.yaml_ then test it with:
```cosole
inv code.test-robot -r robocorp_storage -t "Manage Assets"
```

cmin764 (commented):
Self-review 1

cmin764:
Takes into use `robocorp-storage` **0.3.1** given: https://github.com/robocorp/robo/pull/45

PR #1020, opened 2023-07-24
Use `webdriver_manager` 3.9.1 in `core`

cmin764:
Tested with our current **Selenium** library and it works just fine.

⚠️ Note that the latest **115** _chromedriver_ (required by Google Chrome **115**) may require setting the `binary_location` explicitly into the passed `options` param to the `Open Available Browser` / `Open Browser` keywords.

Test [Process](https://cloud.robocorp.com/orgrobocorp/solutiondemos/processes/bfcf4900-9948-4df8-9cb9-eee7337c5006)
Test [robot](https://robocorp.com/portal/robot/robocorp/example-google-image-search)
Upstream [Issue](https://github.com/SergeyPirogov/webdriver_manager/issues/550)


bogdancondurache (approved):

packages/core/src/RPA/core/webdriver.py
```diff
@@ -5,26 +5,63 @@
...
+#  https://github.com/SergeyPirogov/webdriver_manager/issues/550
+class ChromeDriverManager(_ChromeDriverManager):
+    """Custom Chrome webdriver manager which correctly downloads the chromedriver."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.driver.get_latest_release_version = MethodType(
+            ChromeDriver.get_latest_release_version, self.driver
+        )
```
cmin764, line 62: This transforms an unbound method (attached to class) into a bound method (attached to instance), so we can transplant the new logic without dealing with a new instantiation, which would require the handling of all the input `kwargs` (and their defaults).

cmin764 (commented):
This works like a [charm](https://cloud.robocorp.com/orgrobocorp/solutiondemos/processes/bfcf4900-9948-4df8-9cb9-eee7337c5006/runs/47f7379c-2617-43f7-9eb8-d0c6c223819b/stepRuns/7622440a-c087-49ab-8e8d-1c904f99fe9d/artifacts) within the new Cloud Container, even with Chromium **114**.

PR #1034, opened 2023-07-25
Add support for Selenium binary location through env var

cmin764:
Currently this is needed when developing and/or releasing from a Mac, as for some reason here the new **115** Chrome isn't detected automatically by the _chromedriver_ **115**, but it is by _chromedriver_ **114** even when used with the same Chrome 115.


cmin764:
Working runs with [Process](https://cloud.robocorp.com/orgrobocorp/solutiondemos/processes/bfcf4900-9948-4df8-9cb9-eee7337c5006).

PR #1057, opened 2023-08-10
`RPA.Cloud.Google` Handle "application" type attachments with `MimeApplication` class

mikahanninen:
For example, the **.xslx** are not attached correctly because they are not attached as "application" type.


packages/google/pyproject.toml
cmin764, line 1: Don't forget to change the `black` Python targets as well:
```toml
[tool.black]
target-version = ["py38", "py39", "py310"]
```

I always forgot these :).

cmin764 (commented):
Quick notice during Py 3.7 -> 3.10 replacement.

PR #1068, opened 2023-08-23
`RPA.Assistant` return date object and fix validation

bogdancondurache:


mikahanninen (approved):
LGTM. No test related changes are necessary in relation to this update ?

bogdancondurache:
> LGTM. No test related changes are necessary in relation to this update ?

Unfortunately both changes (fixing the validation and converting user input to a date object) are based on user interaction with the UI, which is entirely skipped by tests at the moment. We have some tests for other areas, more programatic, but only manual ones when it comes to UI interaction. Existing manual tests were used already to check this changes.

PR #1077, opened 2023-08-30
Java element refresh and maximum depth support

### LINKED ISSUE #1060: Optimize Java lib when traversing elements
Relates to: https://github.com/robocorp/java-access-bridge-wrapper/issues/13

cmin764:
Closes #1060 
Depends on: https://github.com/robocorp/java-access-bridge-wrapper/pull/14 (`java-access-bridge-wrapper==1.1.0`)

Ensures an element (like a table) refreshes properly when callbacks are ON and allows a manual refresh to be run when callbacks are not available or turned off. Additionally, adds support for a maximum depth through a parameter limiting how deep the children parsing should go.

### ToDo
- [x] Wrap upcoming JAB wrapper [dependency](https://github.com/robocorp/java-access-bridge-wrapper/pull/14) and take into use the refreshing and maximum depth capabilities:
  - Add a new `Refresh Element` keyword which refreshes just a single element (and its children).
  - Add a `max_depth` param under the library initializer and keywords: `Application Refresh`, `Refresh Element`.
- [x] Refactor `do_action("click")` -> `click()`.
- [x] `Read Table` enhancements:
  - Run an explicit refresh on the element when a table is read. (fixes the actual present children list)
  - Add `visible_only=True` param which if switched to `False`, it will return instead all the existing available children, instead of the screen visible ones only.
- [x] Fix bug with `get_element_actions` (from the upstream) and `set_display_scale_factor` (calls a float instead of assignment). 
- [x] Bump `pywin32` version allowance in **main** and sub-packages.


packages/main/src/RPA/JavaAccessBridge.py
```diff
@@ -420,16 +436,25 @@ class JavaAccessBridge:
...
         by the environment variable ``RC_JAVA_ACCESS_BRIDGE_DLL``.
 
         The ``ignore_callbacks`` can be set to ``True`` if the target application
         does not support Java callback feature (property value change listeners).
 
         :param ignore_callbacks: set to `True` if application does not support
-         Java callback feature, defaults to False
+            Java callback feature, defaults to False
         :param access_bridge_path: absolute filepath to the DLL, defaults to None
+        :param max_depth: limit the height of the element tree (how deep the search goes)
```
mikahanninen, line 455: missing description on the default behaviour on max_depth and disable_refresh

packages/main/src/RPA/JavaAccessBridge.py
```diff
@@ -927,14 +945,27 @@ def toggle_drop_down(self, locator: LocatorType, index: int = 0):
         matching = self._get_matching_element(locator, index)
         matching.toggle_drop_down()
 
+    def _application_refresh(self, force: bool = False):
+        if self.disable_refresh and not force:
+            self.logger.debug("Refresh triggered, but disabled globally by user.")
+            return
+
+        action = "Forcibly" if force else "Automatically"
```
mikahanninen, line 953: unnecessary variable, could be just in place expression like the max depthing

packages/main/pyproject.toml
cmin764, line 1: Just the new features exposed by JAB wrapper, no breaking changes.

packages/main/src/RPA/JavaAccessBridge.py
```diff
@@ -103,7 +98,6 @@ def __init__(
             self.showing = "showing" in self.states
             self.focusable = "focusable" in self.states
             self.node = node
-            self.internal = internal_node
```
cmin764, line 106: This isn't used nor important and creates confusion. (we already have our passed `.node` for that purpose)

packages/main/src/RPA/JavaAccessBridge.py
```diff
@@ -117,7 +111,22 @@ def __init__(
             self.visible_children = node.get_visible_children()
             self.visible_children_count = node.visible_children_count
             self.index_in_parent = node.context_info.indexInParent
-            self.column_count = column_count or len(self.visible_children)
+            self.column_count = column_count
+            self.set_rows_columns(self.visible_children, index=index)
```
cmin764, line 115: Correctly computes the number of columns a row has by checking the cell coordinates. (as soon as the next child will overlap on OX with the first one, it means we're on the 2nd row (if any))

packages/main/src/RPA/JavaAccessBridge.py
```diff
@@ -142,7 +151,7 @@ def type_text(self, text: str, clear: bool = False) -> None:
                 DESKTOP.press_keys("ctrl", "a")
                 time.sleep(0.2)
                 DESKTOP.press_keys("delete")
-            time.sleep(0.2)
+                time.sleep(0.2)
```
cmin764, line 154: Misaligned sleep, we trigger it after pressing the key only.

cmin764 (commented):
Self-review

PR #1088, opened 2023-09-07
Use `core` 11.1.0 (webdriver fixes) and truststore

cmin764:
Release `rpaframework` **27.0.0-alpha**.


packages/main/src/RPA/__init__.py
```diff
@@ -0,0 +1 @@
+import RPA.core
```
cmin764, line 1: @mikahanninen This breaks importing, any ideas for a solution?

PR #1113, opened 2023-10-12
Clarify restrictions on parent/child syntax of Get Elements

kkotenko:
What I would have expected from Windows.Get Elements is that if I provide a locator of the form 

``locator_returning_multiple_elements > locator_returning_multiple_children``, then the return is the union of all applicable children of *all* elements targeted by ``locator_returning_multiple_elements``, similarly to how XPath would process ``locator_returning_multiple_elements[*]``. However, this is not the case - only the first result of ``locator_returning_multiple_elements`` is processed.

This documentation change aims to clarify that. 

(Ideally, that behavior would be changed, but I assume that would be a breaking change...)

### DIFF
--- a/packages/windows/src/RPA/Windows/keywords/locators.py
+++ b/packages/windows/src/RPA/Windows/keywords/locators.py
@@ -176,6 +176,13 @@ class LocatorKeywords(LocatorMethods):
         By default, only the siblings (similar elements on the same level) are taken
         into account. In order to search globally, turn `siblings_only` off, but be
         aware that this will take more time to process.
+
+        Note that if the syntax ``parent_locator > child_locator`` is used
+        in the locator, it is assumed that ``parent_locator`` returns a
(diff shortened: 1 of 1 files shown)


cmin764 (approved):
Indeed, `Get Elements` will default internally to `Get Element` for every parent locator excluding the last leaf, where from there we look into all siblings (or any other different-in-level child) matching the last part of the locator, thus creating this way the final list of element**s**.

So your addition to the documentation is valid @kkotenko, I'll approve the PR, but before merging it we want @mikahanninen opinion as well just in case I missed something.

Thank you!

mikahanninen (approved):

mikahanninen:
Yeah, I think the doc update is good. We could raise an exception if the parent element matches more than 1 element but maybe that is too restrictive.

cmin764:
Merged this, should be available with the next `rpaframework-windows` release.

PR #1128, opened 2023-11-13
Add AES256 encryption

mikahanninen:
```robotframework
*** Settings ***
Library     RPA.Crypto


*** Tasks ***
Fernet with string
    ${fernet_key}=    Generate Key
    Use Encryption Key    ${fernet_key}
    ${token}=    Encrypt string    This is a secret, don't share it
    ${decrypt}=    Decrypt String    ${token}

AES256 with string
    ${aes256_key}=    Generate Key    AES256
    Use Encryption Key    ${aes256_key}    AES256
    ${token}=    Encrypt string    This is a secret, don't share it    encryption_type=AES256
    ${decrypt}=    Decrypt String    ${token}    encryption_type=AES256

Fernet with file
    ${fernet_key}=    Generate Key
    Use Encryption Key    ${fernet_key}
    ${path}=    Encrypt File    example.xlsx
    Log To Console    FERNET Encrypted filepath: ${path}
    Decrypt File    ${path}    output=decrypted_fernet.xlsx    encryption_type=FERNET

AES256 with file
    ${aes256_key}=    Generate Key    AES256
    Use Encryption Key    ${aes256_key}    AES256
    ${path}=    Encrypt File    example.xlsx    encryption_type=AES256
    Log To Console    AES256 Encrypted filepath: ${path}
    Decrypt File    ${path}    output=decrypted_aes256.xlsx    encryption_type=AES256

```


packages/main/src/RPA/Crypto.py
```diff
@@ -104,9 +119,19 @@ def generate_key(self) -> str:
         If the key is lost, the encrypted data can not be recovered.
         If anyone else gains access to it, they can decrypt your data.
         """
-        return Fernet.generate_key().decode("utf-8")
+        if encryption_type == EncryptionType.FERNET:
+            return Fernet.generate_key().decode("utf-8")
+        elif encryption_type == EncryptionType.AES256:
+            key = os.urandom(32)
```
sahokas, line 126: Although `os.urandom` result "should" be good enough for cryptographic operations, i'd be best to use a dedicated cryptographically strong random source.

For example secrets library added in Python 3.6  https://docs.python.org/3/library/secrets.html seems like a good choice for this purpose?

packages/main/src/RPA/Crypto.py
```diff
@@ -344,3 +415,33 @@ def decrypt_file(self, path: str, output: Optional[str] = None) -> str:
...
             return str(output)
+
+    # Helper methods for AES256 (same as before)
+    def _generate_aes256_key(self) -> bytes:
+        key = os.urandom(32)
+        return base64.urlsafe_b64encode(key)
+
+    def _encrypt_aes256(self, data: bytes) -> bytes:
+        backend = default_backend()
+        iv = os.urandom(12)
```
sahokas, line 436: Better use cryptographically strong random also for IV

packages/main/src/RPA/Crypto.py
```diff
@@ -344,3 +415,33 @@ def decrypt_file(self, path: str, output: Optional[str] = None) -> str:
...
+        cipher = Cipher(
+            algorithms.AES(base64.urlsafe_b64decode(self._key)),
+            modes.GCM(iv),
```
sahokas, line 439: GCM mode is a great choice

sahokas (approved):
Looks pretty straightforward. I propose checking the source of random data.

We could also consider adding support for the Additional Authenticated Data (AAD) in AES-GCM. On the other hand, most people would not know what to use it for, and if they know & want to use it, they can implement the encryption themselves fully. So its probably best kept as simple as possible like done now.

PR #1132, opened 2023-11-21
`RPA.Windows` Allow calling `List Windows` without getting process info

mikahanninen:


packages/core/src/RPA/core/windows/helpers.py
```diff
@@ -7,9 +7,12 @@
     import psutil
 
 
-def get_process_list() -> Dict:
+def get_process_list(get_process_info: bool = True) -> Dict:
     """Get process list.
 
     Returns dictionary mapping process id to process name
     """
```
cmin764, line 14: The parameter can be documented in the docstring.

packages/core/src/RPA/core/windows/helpers.py
```diff
@@ -7,9 +7,12 @@
...
-    return {proc.pid: proc.name() for proc in psutil.process_iter()}
+    if get_process_info:
+        return {proc.pid: proc.name() for proc in psutil.process_iter()}
+    else:
+        return {}
```
cmin764, line 18: Since this seems to have no utility when is called with `get_process_info=False` (as it returns an empty dict), maybe is better after all to not control the return content by this parameter and remove it, while controlling the data right in the caller:
```python
process_list = get_process_list() if self.ctx.list_processes else {}
```

packages/core/src/RPA/core/windows/window.py
```diff
@@ -87,10 +89,14 @@ def list_windows(
             rectangle = (
                 [rect.left, rect.top, rect.right, rect.bottom] if rect else [None] * 4
             )
+            if self.ctx.list_processes:
+                name = process_list[pid] if pid in process_list else None
+            else:
+                name = Path(fullpath).name if fullpath else None
```
cmin764, line 95:
```suggestion
            name = process_list[pid] if pid in process_list else None
            if not name and fullpath:
                name = Path(fullpath).name
```
Just in case you feel is better to try to retrieve a `name` from everywhere we can, no matter the circumstances and expectations.

packages/core/src/RPA/core/windows/__init__.py
```diff
@@ -17,6 +17,7 @@ class WindowsElementsMixin:
 
     def __init__(self, locators_path: Optional[str] = None):
         self.logger = logging.getLogger(__name__)
+        self.list_processes = True
```
cmin764, line 20: Seems like a good way to control the library when we're about to decide if we want to list processes by PIDs or not. 

But who/what sets this `list_processes` attribute (as `False` for e.g.) over the library object instance? And do you have a test/example showcasing the usage?

cmin764 (approved):
LGTM, left a couple of comments/questions, but no blocker nor strong opinion on them.

packages/core/src/RPA/core/windows/helpers.py
mikahanninen: It is easier imho to handle return result here than to check all possible places where result is handled and change them.

packages/core/src/RPA/core/windows/window.py
mikahanninen: Yeah maybe but as this is **this** or that **setting** then your suggestion would do 2 operations in the case we really just want to do one.

packages/core/src/RPA/core/windows/helpers.py
mikahanninen: Yeah I can add a docstring.

packages/core/src/RPA/core/windows/helpers.py
cmin764: And wouldn't that require to refactor `get_process_list()` into `get_process_list(self.ctx.list_processes)` in "all possible places where result is handled"?

Also, luckily, this is used just here:
```
cmin@robocorp-mac rpaframework % find . -type d -name src -exec grep -rni "get_process_list" {} \;
./packages/core/src/RPA/core/windows/window.py:10:from RPA.core.windows.helpers import IS_WINDOWS, get_process_list
./packages/core/src/RPA/core/windows/window.py:73:        process_list = get_process_list()
Binary file ./packages/core/src/RPA/core/windows/__pycache__/helpers.cpython-310.pyc matches
Binary file ./packages/core/src/RPA/core/windows/__pycache__/window.cpython-310.pyc matches
./packages/core/src/RPA/core/windows/helpers.py:10:def get_process_list() -> Dict:
Binary file ./packages/windows/src/RPA/Windows/__pycache__/utils.cpython-310.pyc matches
./packages/windows/src/RPA/Windows/utils.py:4:from RPA.core.windows.helpers import IS_WINDOWS, get_process_list  # noqa: F401
```

mikahanninen: The `get_process_list()` will be having a default setting so there is no need to give the value for that.

packages/core/src/RPA/core/windows/__init__.py
mikahanninen: Was planning to adding keyword to Windows library to toggle this setting but obviously it requires first that this value exists in the `core`.

packages/core/src/RPA/core/windows/window.py
mikahanninen: Ah good point. I will change this then.

packages/core/src/RPA/core/windows/helpers.py
mikahanninen: I was not sure if this is used only in the rpaframework packages or maybe even directly by the users / in inspector or in whatever place.

PR #1147, opened 2024-01-30
Bump Robotframework support to <8.0.0

lugi0:
This PR simply allows rpaframework to run alongside Robot Framework 7, but it has not been tested for compatibility with it.
There's also a version bump to pylint (which was roughly 2 years out of date) in order to resolve a dependency failure caused by the `wrapt` library not working correctly with Python 3.11 until version 1.14.0


pkraju27:
Do we have any ETA for this Change to be available.

NorseGaud:
Bump

lucasvannoort:
@lugi0 is it possible to merge this PR please ?
I had to upgrade my Robot Framework project to the latest version 7.1.1...

lugi0:
@lucasvannoort I am not a maintainer for this repo, it's not up to me. 
When I tested it locally this PR worked fine for me, so if you need to you could just use these changes in a local copy of the library and use that instead of the official one.

lhlhai:
Does anyone have the current status of this PR?

mikahanninen:
Sorry, but I had to implement this without this PR to sync all packages to the same level while also improving internal tooling within the repository.

Thank you for the PR anyway @lugi0

PR #1217, opened 2025-03-20
Upgrade actions/upload-artifact to v4

kkotenko:
Currently, the "daily installation tests" github action is disabled because it no longer works (see e.g. https://github.com/robocorp/rpaframework/actions/runs/13963050710 )

I *think* this PR restores the desired behavior, per https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#overwriting-an-artifact (linked from the failure notification). If not, guidance can be found in that document.


mikahanninen:
Thank you for the PR ❤️

PR #1219, opened 2025-03-20
upgrade installation checks for mac and windows to python 3.11.9

kkotenko:
Upgrades the macos and windows runners to 3.11.9 (most current available version of 3.11 according to https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json )

Successful run with this config is available at https://github.com/kkotenko/rpaframework/actions/runs/13967270241


mikahanninen:
This should be now in better state so this PR is not needed, but thank you anyways ❤️

